PR #5, opened 2018-01-19
public api basic interface + initial hard-coded smart contract support + first simple e2e test

idandash:
## Major Changes & Additions: 
1. Public API service exposes _sendTransaction()_ and _call()_ methods (via a gRPC interface).
2. Virtual machine service implements an hard-coded contract processor, and auto-registers a test smart contract named "Bar Smart Contract" that allows holding and transferring a test cryptocurrency named Bar.
3. "Simple token transfer" end-to-end test.

### The End-to-End Test

- Bar is our test cryptocurrency represented through an hard-coded smart contract named "Bar Smart Contract".
- Our test aims to test a simple transfer of Bars between two accounts.
- The test initializes accounts A and B with a number of Bars. Then, transfers Bars from account A to account B. After each stage, it checks, within a few seconds, that the balances of the affected accounts have changed properly.
- Both the balance initialization and the transfer are done by sending a transaction to the network through a node (a _sendTransaction()_ request), while the balance check is done by querying a trusted node for the balance (a _call()_ request).


.gitignore
```diff
@@ -22,6 +22,8 @@ bower_components
 
 # Editors
 .idea
+.history
+.vscode
```
lbeder: No biggy, but `.vscode` appears twice.

e2e/src/simple-token-transfer.spec.ts
```diff
@@ -0,0 +1,172 @@
+const {Assertion, expect} = require("chai");
```
lbeder: What about creating a new `tests` folder and move `e2e` there (in addition to per-project unit tests we'd add soon)? We're very (very) soon have other "global" test tools and projects there.

e2e/src/simple-token-transfer.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+        return this.sendTransaction("transfer", [to, amount]);
+    }
+
+    public getMyBalance() {
+        return this.call("getMyBalance", []);
+    }
+}
+
+class OrbsClient {
+    crypto: CryptoUtils;
```
lbeder: It reminds me: I'll make sure that our CryptoUtils is a "native" library, supporting everything we might need (probably ed25519, SHA3, RIPEMD160 for now).

e2e/src/simple-token-transfer.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+    accounts.set(account.getAddress(), account);
+
+    if (input.bars > 0)
+        await account.initBarBalance(input.bars);
+
+    return account;
+}
+
+describe("simple token transfer", function() {
+    this.timeout(400000);
```
lbeder: Having these kind of timeouts in tests won't "scale" very soon and kill us. Let's think together of how can we make these tests more deterministic (e.g., polling or some way of stubbing).

e2e/src/simple-token-transfer.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    public getAddress() {
+        return this.crypto.getPublicKey();
+    }
+
+}
+
+
+async function assertModelBars (n: number) {
```
lbeder: Nice :)

projects/public-api-service-typescript/src/service.ts
```diff
@@ -25,32 +25,54 @@ export default class PublicApiService {
     this.askForHeartbeat(this.peers.gossip);
   }
 
-  async generateRandomTransaction() {
-    const senderCrypto: CryptoUtils = CryptoUtils.initializeTestCrypto(`user${Math.floor((Math.random() * 10) + 1)}`);
-    const sender = senderCrypto.getPublicKey();
+  // async generateRandomTransaction() {
```
lbeder: Do we still need it? If not, then let's just delete the commented out code.

projects/public-api-service-typescript/src/service.ts
```diff
@@ -25,32 +25,54 @@ export default class PublicApiService {
...
+    rpc.res = {
+      resultJson: resultJson
+    };
   }
 
   async main() {
     this.peers = topologyPeers(topology.peers);
     setInterval(() => this.askForHeartbeats(), 5000);
-    setInterval(() => this.generateRandomTransaction(), Math.ceil(Math.random() * 30000));
+    // setInterval(() => this.generateRandomTransaction(), Math.ceil(Math.random() * 30000));
```
lbeder: ^ same like above

projects/virtual-machine-service-typescript/src/hard-coded-contracts/processor.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+export default class HardCodedSmartContractProcessor {
+    contractAddresses = new Map<string, any>();
+    stateStorageClient: types.StateStorageClient;
+
+    constructor(stateStorageClient: types.StateStorageClient) {
+        this.stateStorageClient = stateStorageClient;
+
+        this.registerContract("bar-smart-contract", "barbar");
```
lbeder: This doesn't feel outside of the testing suite, but I'm sure we will refactor it soon enough.

projects/virtual-machine-service-typescript/src/hard-coded-contracts/registry/bar-smart-contract.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export default class BarSmartContract extends BaseSmartContract {
+    static readonly SUPERUSER = "0000";
+
+    public async transfer(recipient: string, amount: number) {
+        if (amount === 0) {
+            throw new Error("transaction amount must be > 0");
+        }
+
+        const senderBalance: number = await this.getBalance(this.sender);
```
lbeder: We should start using `BigNumber` only, in these cases. The token amounts and denominations we will be working with won't fit the `number` resolution.

projects/virtual-machine-service-typescript/src/hard-coded-contracts/registry/bar-smart-contract.ts
```diff
@@ -0,0 +1,48 @@
...
+        if (senderBalance < amount) {
+            throw new Error(`balance is not sufficient ${senderBalance} < ${amount}`);
+        }
+
+        // TODO: no integer overflow protection
```
lbeder: Ah, I see that you've noticed it yourself as well :D.

projects/virtual-machine-service-typescript/src/hard-coded-contracts/registry/bar-smart-contract.ts
```diff
@@ -0,0 +1,48 @@
...
+    private async setBalance(account: string, amount: number) {
+        return this.stateAccessor.store(this.getAccountBalanceKey(account), JSON.stringify(amount));
+    }
+
+    private getAccountBalanceKey(account: string) {
+        return `balances.${account}`;
+    }
+
+
+}
```
lbeder: [insignificant]: Some style issues (new lines, commented code, `number  = ` instead of `number =`, etc.

lbeder (approved):
Awesome stuff 👍!

I've left some comments, but nothing critical. Take a look and see if you feel that some should make this PR as well (e.g., `BigNumber` stuff, new `tests` folder, etc.).

In general, I don't like that we're having test hard-coded contracts outside of a testing suite, but it's totally understandable for now - we'd just need to think how to restructure/refactor this in the future.

e2e/src/simple-token-transfer.spec.ts
electricmonk: In general I'd avoid naming anything `FooUtils` or `CommonFoo`, as it lends legitimacy to these pieces of code becoming rubbish bins. Try to decide what does it do and name it accordingly. 

See here: http://arlobelshee.com/good-naming-is-a-process-not-a-single-step/

e2e/src/simple-token-transfer.spec.ts
electricmonk: Agreed. Ideally the test will use the same update mechanism as the client application (a Wallet, for instance). Maybe we need to expose a websocket API for these updates, for instance.

projects/virtual-machine-service-typescript/src/hard-coded-contracts/processor.ts
electricmonk: @lbeder you mean doesn't feel right?

projects/virtual-machine-service-typescript/src/hard-coded-contracts/registry/bar-smart-contract.ts
electricmonk: if BigNumber is like Java's BigDecimal, then it's also mandatory due to precision and rounding issues

e2e/src/simple-token-transfer.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+describe("simple token transfer", function() {
+    this.timeout(400000);
+    it("transfers 1 token from one account to another", async function() {
+        this.timeout(400000);
+
+        console.log("initing account1 with 2 bars");
+        const account1 = await anAccountWith({bars: 2});
+        console.log("initing account2 with 0 bars");
+        const account2 = await anAccountWith({bars: 0});
+        await delay(5000);
```
electricmonk: this is actually much worse IMO than the timeout. It's better to use an `eventually` assertion as provided by Chai As Promised http://chaijs.com/plugins/chai-as-promised/

projects/virtual-machine-service-typescript/src/hard-coded-contracts/processor.ts
lbeder: Yeah...

projects/virtual-machine-service-typescript/src/hard-coded-contracts/registry/bar-smart-contract.ts
lbeder: I agree. IMHO, we should switch to use only BigNumber's before it'd be too late.

e2e/src/simple-token-transfer.spec.ts
idandush: @electricmonk WDYT?

e2e/src/simple-token-transfer.spec.ts
idandush: Agreed, but it's not related to this specific pull request.

projects/virtual-machine-service-typescript/src/hard-coded-contracts/processor.ts
idandush: We'll do. It's just that it's not currently possible to register contracts externally to the virtual machine service (they are hard-coded and pre-loaded).

projects/virtual-machine-service-typescript/src/hard-coded-contracts/registry/bar-smart-contract.ts
idandush: We're not using BigNumber (yet) since we need _protobuf_ and _grpc_ to support it. There's a separate task on that and it will hopefully be solved in the next few days. I've already started exploring it.

PR #6, opened 2018-01-22
Add docker & run integration test inside the container

netoneko:
This PR does not deal with splitting up `simulate.js`, but it's in the pipeline down the road.

What can we do now:
- [x] Run `docker-compose up` and see a working docker image
- [x] Run integration test (`e2e`) inside the image

screen shot 2018-01-22 at 11 55 09

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,7 @@
 FROM node:9-alpine
 
-RUN apk --no-cache add bash python build-base
+RUN apk --no-cache add bash git
+# python build-base
 # remove python and build-base when grpc-prebuilt will be up again
 
 ADD . /opt/orbs
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+FROM node:9-alpine
+
+RUN apk --no-cache add bash python build-base
+# remove python and build-base when grpc-prebuilt will be up again
+
+ADD . /opt/orbs
+
+WORKDIR /opt/orbs
+
+RUN npm install -g typescript@2.6.2
+
+RUN ./install.sh
+
+RUN find . -name package-lock.json -delete
+
+RUN ./build.sh
+
+CMD node simulate/src/index.js transaction-gossip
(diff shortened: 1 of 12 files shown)


Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+
+RUN npm install -g typescript@2.6.2
+
+RUN ./install.sh
+
+RUN find . -name package-lock.json -delete
+
+RUN ./build.sh
+
+RUN cd e2e && ./build.sh
```
lbeder: Should we add `e2e` as part of the main `build.sh`?

config/projects.json
```diff
@@ -45,5 +45,6 @@
     "runtime": "typescript",
     "service": "virtual-machine",
     "type": "service"
-  }
+  },
+  "order": ["architecture", "common-library-typescript", "block-storage-service-typescript", "chatter-service-typescript", "consensus-service-typescript", "gossip-service-typescript", "public-api-service-typescript", "state-storage-service-typescript", "transaction-pool-service-typescript", "virtual-machine-service-typescript"]
```
lbeder: Can you break it to multilines, for readability?

projects/common-library-typescript/build.sh
```diff
@@ -1,3 +1,4 @@
 npm link ../architecture
+npm install --unsafe-perm grpc@1.8.4
```
lbeder: Can you add a comment of why it was required, so that we won't forget?

Dockerfile
netoneko: We could, but at some point we're going to exclude it anyway, and probably have a different Dockerfile for production builds.

lbeder: Makes sense, yeah.

projects/common-library-typescript/build.sh
netoneko: Done.

config/projects.json
netoneko: Done.

PR #9, opened 2018-01-22
Initial support for AWS ECS

netoneko:
Run `orbs-network` container on ECS.

Next step would be scaling and discovery in a common network with ability to scale up and down.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ jobs:
    machine: true
    steps:
      - checkout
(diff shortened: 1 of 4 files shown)


idandash:
@netoneko AWS account-specific settings should not be hard-coded

netoneko:
@idandash fixed.

lbeder:
@netoneko do you feel that this PR is good to go?

netoneko:
@lbeder yes. It paves the road to next steps of running our infrastructure both locally and in the cloud.

PR #11, opened 2018-01-22
Initial SidechainConnector service w/ Ethereum call() support

idandash:


projects/architecture/interfaces/sidechain-connector.proto
```diff
@@ -0,0 +1,34 @@
...
+    string name = 1;
+    repeated EthereumFunctionParameter inputs = 2;
+    repeated EthereumFunctionParameter outputs = 3;
+}
+
+message CallEthereumContractInput {
+  string contractAddress = 1;
+  EthereumFunctionInterface functionInterface = 2;
+  repeated string parameters = 3;
+// string blockNumber = 5;
```
lbeder: Why did you want it? In order to call the contract in a "specific time"?

projects/sidechain-connector-service-typescript/build.sh
```diff
@@ -0,0 +1,3 @@
+npm link ../common-library-typescript
```
lbeder: @idandash could you decorate the bash files according to what I've added last week? I know that you've added these prior to my change, but this way we could stay in-sync.

projects/sidechain-connector-service-typescript/src/index.ts
```diff
@@ -0,0 +1,9 @@
+import { topology, grpc } from "orbs-common-library";
+import SidechainConnectorService, { SidechainConnectorServiceOptions } from "./service";
+
+const server = grpc.sidechainConnectorServer({
+  endpoint: topology.endpoint,
+  service: new SidechainConnectorService({
+    ethereumNodeHttpAddress: process.argv[3] // argv[2] is "taken" by orbs-common-library/topology.ts
```
lbeder: What about adding a more convenient ARGV parser? Personally, I like `yargs` (it also sounds like a pirate!).

projects/sidechain-connector-service-typescript/src/service.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  async askForHeartbeat(peer: types.HeardbeatClient) {
+    const res = await peer.getHeartbeat({ requesterName: topology.name, requesterVersion: topology.version });
+    logger.info(`${topology.name}: received heartbeat from '${res.responderName}(v${res.responderVersion})'`);
+  }
+
+  askForHeartbeats() {
+  }
+
+
```
lbeder: Eventually, we'd need to remove these style errors from everywhere 😅

projects/sidechain-connector-service-typescript/src/test/ethereum-connector.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+const Web3 = require("web3");
+const ganache = require("ganache-cli");
+const path = require("path");
+const fs = require("fs");
+const solc = require("solc");
+
+import EthereumConnector from "../ethereum-connector";
+
+const SIMPLE_STORAGE_SOLIDITY_CONTRACT = `
+pragma solidity ^0.4.17;
```
lbeder: Thinking out-loud: eventually, we might find ourselves calling contracts which were compiled with a different solidity compiler (which can result in a different and incompatible API). Should we support it (outside the tests, that is)? E.g., by adding a compiler version to the message?

projects/sidechain-connector-service-typescript/src/types/web3.ts
```diff
@@ -0,0 +1,401 @@
+import { BigNumber } from "bignumber.js"; // TODO change to BN
```
lbeder: Where is it from? AFAIK, there should be official types soon, but why not to use something like https://github.com/slockit/web3.ts for now (there are others as well)?

projects/sidechain-connector-service-typescript/src/types/web3.ts
```diff
@@ -0,0 +1,401 @@
...
+import * as us from "underscore";
```
lbeder: What about using `lodash` instead? It's better and more optimized than `underscore` and AFAIK, we're already using it elsewhere.

projects/sidechain-connector-service-typescript/src/index.ts
netoneko: I'd rather have it in `topology.ts` so I can maybe pass it as an environment variable later.

lbeder: @netoneko I've also addresses this in the general comment below. We need to think of how are we going to deploy an Ethereum full node instance/pod and then bind and "discover" them alongside the Ethereum connector service (which will be a docker of its own, eventually). Let's discuss together with @idandash tomorrow.

lbeder:
@idandash I'm quickly making some of the above suggestions myself and then merging 👍

projects/architecture/interfaces/sidechain-connector.proto
idandash: Yes, but it's not a must at the moment

projects/sidechain-connector-service-typescript/src/types/web3.ts
idandash: it's an external lib. I haven't touched it

projects/sidechain-connector-service-typescript/src/test/ethereum-connector.spec.ts
idandash: Eventually yes. it's not a requirement at the moment

PR #17, opened 2018-01-27
New consensus + some refactoring and cleanup

lbeder:
In this PR:

1. For our testnet, I've replaced the dummy PBFT we had with a RAFT-like consensus protocol, based on the https://github.com/ben-ng/gaggle (which I've extended for our needs here: https://github.com/lbeder/gaggle). I've spent a lot of time in evaluating existing consensus implementations (mostly focusing on JavaScript implementations, but also played with golang and Python) and, at the moment, this is the best implementation I've found and should help us bootstrapping the testnet. For the mainnet, we will replace this implementation with Orbs Consensus Protocol.

2. I've added an automatic uncaught exception and rejections handing via the `ErrorHandler` and added it to our services. Check it out.

3. Some style and cleanup on the way.

P.S.,
I'm leaving the current dummy PBFT for a while, until we'll feel well with this implementation and then I'd remove it as well.


lbeder:
@idandash @netoneko please review, but don't merge it yet. The tests are passing, but I'd prefer to see them passing on the CI first (which will be possible once @netoneko's fix will get merged).

netoneko:
@lbeder is there any configuration option that allows us to switch between new consensus and old consensus?

lbeder:
@netoneko not at the moment, since I'm aiming to delete the old implementation as soon as possible.

lbeder:
@idandash @netoneko resolved (new) conflicts, so the branch is mergeable again. Since it's currently blocking @netoneko PR (since it has some "leader-ful" stuff), better to merge sometime tomorrow. I'll continue working on my consensus backlog on top of a new feature branch.

PR #19, opened 2018-01-29
CloudFormation template and bootstrap script for a new node

netoneko:
This PR depends on #18. Essentially, it allows us to bootstrap new node and add it to the network (assuming we have configured all ip addresses beforehand).

- [x] CloudFormation template
- [x] Bootstrap script
- [x] Dockerize EVERYTHING
- [x] Run e2e on the cluster

There are some AWS account-specific things, I will get rid of them in the future.

screen shot 2018-01-31 at 10 59 12


config/staging.json
```diff
@@ -0,0 +1,19 @@
...
+    "level": "debug",
+    "fileName": "logs/staging.log",
+    "maxSize": 1024000,
+    "maxFiles": 4,
+    "logzio": {
+      "enabled": false
+    }
+  },
+  "consensus": {
+    "clusterSize": 7,
```
lbeder: @netoneko just to double check - are we running 7 nodes on staging indeed? Not 6, like via `simulate.sh`?

deploy/bootstrap/.env
```diff
@@ -0,0 +1,2 @@
+GOSSIP_PEERS=ws://bliny.nodes.orbs-test.com:60001,ws://pelmeni.nodes.orbs-test.com:60001,ws://borscht.nodes.orbs-test.com:60001,ws://pirogi.nodes.orbs-test.com:60001,ws://oladyi.nodes.orbs-test.com:60001,ws://olivier.nodes.orbs-test.com:60001,ws://solyanka.nodes.orbs-test.com:60001
```
lbeder: Can you write these as multilines, for improved readability?

deploy/bootstrap/docker-compose.yml
```diff
@@ -0,0 +1,66 @@
+version: '2'
+
+services:
+  consensus:
+    image: ${DOCKER_IMAGE}:${DOCKER_TAG}
+    working_dir: /opt/orbs/projects/consensus-service-typescript/
+    command: node dist/index.js /opt/orbs/config/topologies/discovery/node1/consensus.js
```
lbeder: Why is it `node1` everywhere?

deploy/cloudformation/create-stacks.js
```diff
@@ -0,0 +1,59 @@
...
+
+const deployAll = () => {
+    return Promise.all(stacks.map(deploy));
+}
+
+if (process.argv[2] === '--remove-all') {
+    removeAll().then(console.log, console.error);
+} else {
+    deployAll().then(console.log, console.error);
+}
```
lbeder: It's not important here, but let's try to use TypeScript as much as possible, so that we'd gain some testability and standardization for our "internal" scripts as well.

deploy/cloudformation/create-stacks.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
lbeder: Can you decorate it with `-e` like with other shell files?

deploy/cloudformation/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "cloudformation",
+  "version": "1.0.0",
```
lbeder: Let's start with something like `0.1.0` for now.

deploy/cloudformation/package.json
```diff
@@ -0,0 +1,14 @@
...
+  "description": "## Template and parameters",
+  "main": "create-stacks.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
lbeder: You have tests, though. Why not to run them via `npm test`?

deploy/cloudformation/parameters.json
```diff
@@ -0,0 +1,62 @@
+[
+    {
+        "ParameterKey": "KeyName",
+        "ParameterValue": "kirill"
```
lbeder: Hi Kiriil! ;)

(Can we change to something less... personal)?

deploy/cloudformation/test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
lbeder: Same like above.

config/staging.json
netoneko: Yes, we are running 7 nodes.

deploy/bootstrap/.env
netoneko: I'm not sure it's a very good idea for an `.env` file, we might want to parse it somewhere from scratch and newlines might not be supported.

deploy/bootstrap/docker-compose.yml
netoneko: Because it's a single configuration suitable for any node.

deploy/cloudformation/create-stacks.sh
netoneko: Sure.

deploy/cloudformation/package.json
netoneko: It's a shorthand for remote e2e test, not a test for package itself.

deploy/cloudformation/parameters.json
netoneko: Sure.

deploy/bootstrap/.env
lbeder: Alright.

deploy/bootstrap/docker-compose.yml
lbeder: Ah, I see. Perhaps we will rename it to `node` in the future.

deploy/cloudformation/create-stacks.js
lbeder: That's a valid point. Never mind for now.

PR #24, opened 2018-02-01
Feature/basic key management

netoneko:
Dummy keys are stored on S3 and downloaded on node bootstrap.

- [x] update bootstrap script
- [x] update docker-compose.yml
- [x] filter out self from possible peers


idandash:
@netoneko Please rebase with _master_. The e2e test has been completely refactored. You no longer need to use E2E_PUBLIC_API_ENDPOINT etc.

netoneko:
@idandash I base my branch on master, and I had to add `E2E_PUBLIC_API_ENDPOINT` back, because topology dependency is unnecessary for running tests in the cloud.

PR #26, opened 2018-02-01
Add logzio to staging

netoneko:
- [x] support logz.io
- [ ] use conditional imports

screen shot 2018-02-01 at 17 40 30


projects/common-library-typescript/src/logger.ts
```diff
@@ -3,9 +3,11 @@ import "source-map-support/register";
 
 import { config } from "./config";
 import * as winston from "winston";
-import * as winstonLogzioTransport from "winston-logzio";
 import * as path from "path";
 import * as fs from "fs";
+const winstonLogzioTransport = require("winston-logzio");
```
lbeder: Can you use a dynamic import instead?

projects/common-library-typescript/src/logger.ts
```diff
@@ -42,8 +44,17 @@ export class Logger {
       ]
     });
 
+    this._logger.rewriters.push(function(level, msg, meta) {
```
lbeder: I think that this can be done in a cleaner way using its initializer.

projects/common-library-typescript/src/logger.ts
```diff
@@ -121,9 +132,10 @@ export class Logger {
       throw new Error("Missing API key!");
     }
 
-    this._logger.add(winston.transports.Logzio, {
+    this._logger.add(winstonLogzioTransport, {
```
lbeder: See above.

projects/common-library-typescript/src/logger.ts
```diff
@@ -121,9 +132,10 @@ export class Logger {
...
       token: apiKey,
       host: Logger.LOGZIO_HOST,
+
```
lbeder: Empty line.

lbeder:
@netoneko https://github.com/orbs-network/orbs-network/pull/29 relates to some of what you've done here.

config/staging.json
```diff
@@ -4,6 +4,7 @@
     "fileName": "logs/staging.log",
     "maxSize": 1024000,
     "maxFiles": 4,
+    "console": true,
```
lbeder: This one should be `false`, no?

netoneko: No, I did it on purpose, so I could do `docker logs -f` and see the results.

lbeder: Ah, got it.

projects/common-library-typescript/src/logger.ts
lbeder: 👍

lbeder:
I'm waiting for the tests to pass and then merging.

PR #30, opened 2018-02-04
minor code improvements in e2e script

idandash:


e2e/src/simple-token-transfer.spec.ts
```diff
@@ -65,21 +43,37 @@ if (nconf.get("E2E_NO_DEPLOY")) {
     publicApiClient = testEnvironment.getPublicApiClient();
 }
 
+async function assertFooBarAccountBalance (n: number) {
+    // make sure we are working with am Account model
+    new Assertion(this._obj).to.be.instanceof(FooBarAccount);
```
lbeder: Personally, I'd rather to stick to `expect`, e.g.: `expect(this._obj).to.be.an.instanceof(FooBarAccount);`

PR #34, opened 2018-02-05
Block storage

lbeder:
In this PR:

1. Introducing persistency to our persistent block storage.
2. Some cleanup and refactoring.


projects/block-storage-service-typescript/src/block-storage.ts
```diff
@@ -0,0 +1,141 @@
...
+import * as levelup from "levelup";
+import leveldown from "leveldown";
+
+import { logger, config, types, topology, topologyPeers, CryptoUtils, Encoding } from "orbs-common-library";
+
+const crypto = CryptoUtils.loadFromConfiguration();
+
+export default class BlockStorage {
+  // The path to the local LevelDB will be of the format: "../../db/[ENVIRONMENT]/blocks_[NODE].db".
+  public static readonly LEVELDB_NAME: string = `blocks_${crypto.whoAmI()}.db`;
```
netoneko: Why do we need different file names per node?

projects/block-storage-service-typescript/src/block-storage.ts
```diff
@@ -0,0 +1,141 @@
...
+  public static readonly LEVELDB_PATH: string = path.resolve(path.join("../../", "db", config.getEnvironment(),
```
netoneko: Why is path dependent on the environment?

projects/block-storage-service-typescript/src/block-storage.ts
lbeder: We don't. Good point.

lbeder:
@netoneko node/env specific names will go away, when e2e runs in a fully dockerized set up as well.

lbeder:
@idandash I've added the refactoring we've discussed.

lbeder:
@netoneko perhaps that error is because there is no volume for the SB `VOLUME [ "/opt/orbs/logs" ]

lbeder:
@netoneko @idandash don't merge yet - I'm removing the "crypto dependency".

lbeder:
@netoneko @idandash there were some issues with the dockerized tests, so with @netoneko's help - I was able to resolve them. In addition to that, I've allowed myself to correct the local docker tests,  remove `simulate` and friends, and update the README.md.

For all sakes and purposes, this PR is complete.

projects/block-storage-service-typescript/src/block-storage.ts
```diff
@@ -0,0 +1,101 @@
+import * as path from "path";
+
+import { logger, types } from "orbs-common-library";
+
+import LevelDBDriver from "./leveldb-driver";
+
+export default class BlockStorage {
+  public static readonly LEVELDB_PATH: string = path.resolve("../../db/blocks.db");
```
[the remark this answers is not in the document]
lbeder: No, the drive is somewhat an implementation detail (it's not a pure, general purpose driver yet). Let me move it to be a constructor parameter.

PR #36, opened 2018-02-06
Some package fixes

lbeder:


netoneko:
I think we have typescript installed globally somewhere in the build script, please update it too.

lbeder:
@netoneko done

PR #38, opened 2018-02-06
rolled back implementation of configurable block storage path

idandash:


netoneko:
Why did you have to roll it back? I suggest we just introduce a default value.

lbeder:
@netoneko it's a long story, but we've decided to keep the interface as before. In any case, we'll iterate.

PR #45, opened 2018-02-07
Introducing shorter feed back cycles for testing

netoneko:
To achieve shorter feedback cycles, do the following:

1. Run `./docker-build.sh`, it will produce a new docker image
2. Start `./watch.sh`
3. Now `LOCAL=true ./docker-test.sh` will run e2e test on your configuration with your changes automatically applied.

If you want nodes to stay up, use `STAY_UP=true ./docker-test.sh`.

All the logs are in `./logs/$NODE_NAME/$SERVICE/test.log`

Shared volumes mean that you don't have to rebuild your docker image all the time to see the changes, only if you add new libraries or switch to a different branch (this one just to be safe).


netoneko:
@lbeder I have resolved some issues:

1. From my experience, you don't really need `./build.sh` unless you're missing libraries
2. Uncommenting local settings is now encapsulated by `LOCAL` variable.

PR #47, opened 2018-02-07
added my fork of grpc-caller

electricmonk:


projects/common-library-typescript/package.json
```diff
@@ -20,7 +20,7 @@
   "dependencies": {
     "bs58": "^4.0.1",
     "grpc": "^1.8.4",
-    "grpc-caller": "^0.4.0",
+    "grpc-caller": "https://github.com/electricmonk/grpc-caller",
```
idandash: Much safer to enforce a specific version (aka git version tag)

electricmonk: why? it's my fork

electricmonk:
@lbeder not sure what you mean by using `Metadata`; what I aim to achieve here is to have the client add extra data to the error and to alter its message. `Metadata` seems to be a way to transfer context from client to server, i.e. authentication or other stuff that you would want to propagate between servers

projects/common-library-typescript/package.json
idandash: Sure. but even your fork might be updated at some point.

electricmonk: whereupon I'll want the project to get the most recent version, no?

lbeder:
@electricmonk it feels too excessive to me. Why can't we add this logic externally, by wrapping our `grpc` abstraction layer with `try/catch`? Forking such a major package, just for that, thus forcing us to start syncing it explicitly with the mainstream. feels like an overkill.

electricmonk:
Dude. It’s not grpc. It’s a small wrapper.

Shai
________________________________

PR #61, opened 2018-02-12
added transaction handler to public API

electricmonk:
added transaction handler mechanical aspects to the public API lib;
currently the feature is disabled via hard-coded feature toggle until we can run Ethereum connector in e2e test

### DIFF
--- a/projects/libs/core-library-typescript/package.json
+++ b/projects/libs/core-library-typescript/package.json
@@ -20,7 +20,6 @@
   },
   "dependencies": {
     "@types/bluebird": "^3.5.20",
-    "@types/mocha": "^2.2.48",
     "bignumber.js": "^6.0.0",
     "bluebird": "^3.5.1",
     "bs58": "^4.0.1",
@@ -48,6 +47,10 @@
   "devDependencies": {
     "@types/bs58": "^3.0.30",
     "@types/chai": "^4.1.2",
+    "@types/chai-as-promised": "^7.1.0",
+    "@types/mocha": "^2.2.48",
+    "@types/sinon": "^4.1.3",
+    "@types/sinon-chai": "^2.7.29",
     "@types/crypto-js": "^3.1.38",
     "@types/lodash": "^4.14.99",
     "@types/mkdirp": "^0.5.2",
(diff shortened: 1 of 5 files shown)


projects/libs/core-library-typescript/package.json
```diff
@@ -57,9 +57,11 @@
     "@types/winston": "^2.3.7",
     "@types/ws": "^4.0.0",
     "chai": "^4.1.2",
+    "chai-as-promised": "^7.1.1",
```
lbeder: Could you add `@types/chai-as-promised` and `@types/sinon-chai` (or are they already included with something else?)

projects/libs/core-library-typescript/src/public-api/transaction-handler.ts
```diff
@@ -0,0 +1,30 @@
+import { types } from "../common-library/types";
+
+export interface TransactionHandlerConfig {
+    validateSubscription(): boolean;
```
lbeder: Can you change the indentation to 2 spaces, like in (most of the) other files?

projects/services/public-api-service-typescript/src/service.ts
```diff
@@ -3,14 +3,21 @@ import bind from "bind-decorator";
 
 import { logger, config, topologyPeers, grpc, types } from "orbs-core-library/dist/common-library";
 import { topology } from "orbs-core-library/dist/common-library/topology";
-
+import { TransactionHandler, TransactionHandlerConfig } from "orbs-core-library/dist/public-api//transaction-handler";
```
lbeder: I'm removing the `dist` in my PR, but keep this in your code and I'll remove it while merging mine.

projects/libs/core-library-typescript/package.json
electricmonk: shit, I forgot to port them from the other PR

projects/libs/core-library-typescript/src/public-api/transaction-handler.ts
electricmonk: don't we have an editor config file? this should be automatic

lbeder: We don't have one. You are right, it's should be added and automatic.

PR #62, opened 2018-02-12
.editorconfig and recommended extensions

electricmonk:
This PR adds two quasi-related files:
* `.editorconfig` file for cross-project editor settings such as ident size
* .vscode/extensions.json, which is deliberately exempt from `.gitignore` so that new team members could install all extensions we use automatically

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+
+root = true
+
+[*]
+trim_trailing_whitespace = true
+insert_final_newline = true
+end_of_line = lf
+charset = utf-8
+max_line_length = 180
+
+[*.ts]
+indent_style = space
+indent_size = 2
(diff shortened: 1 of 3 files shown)


.editorconfig
```diff
@@ -0,0 +1,13 @@
+
+root = true
+
+[*]
+trim_trailing_whitespace = true
+insert_final_newline = true
+end_of_line = lf
+charset = utf-8
+max_line_length = 180
```
lbeder: This is exactly right for our displays 😂

electricmonk: well, I dunno about you but my monitor is split between IDE and browser. Also, needs to fit on a laptop screen.

lbeder: No, I was joking :). `180` is totally fine. I usually go into flamewars with people who insist on `80`.

PR #71, opened 2018-02-18
Base service

lbeder:
Refactor and DRY using a common base service class, specially:

1. Services can be run using a `ServiceRunner`, which abstracts its async startup (and later - termination) via the `initialize` method.
2. Heartbeat sending, receiving, and registering logic is now common to all services.
3. RPC methods need to be decorated as `@RPCMethod` (or `@SilentRPCMethod` if no logging is desired), which automatically adds the @bind decorator and method logging. 
4. Improved synchronization between multi-proto services.

TBD (later): state machine management.


projects/libs/core-library-typescript/src/base-service/service.ts
```diff
@@ -0,0 +1,72 @@
...
+  protected static RPCMethod(target: Object, propertyKey: string, descriptor: TypedPropertyDescriptor<Function>,
+    silent: boolean = false): any {
+    if (!descriptor || (typeof descriptor.value !== "function")) {
+      throw new TypeError(`Only methods can be decorated with @RPCMethod. <${propertyKey}> is not a method!`);
+    }
+
+    if (!silent) {
+      const originalMethod = descriptor.value;
+      descriptor.value = function(rpc: any) {
+        logger.debug(`${this.nodeTopology.name}: ${propertyKey} ${JSON.stringify(rpc.req)}`);
```
netoneko: Please log it as a message and an object so we could group data in Kibana.

projects/libs/core-library-typescript/src/base-service/service.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  async askForHeartbeat(peer: types.HeardbeatClient) {
+    const res = await peer.getHeartbeat({ requesterName: this.nodeTopology.name, requesterVersion: this.nodeTopology.version });
+
+    logger.debug(`${this.nodeTopology.name} (${this.name}): received heartbeat from '${res.responderName}(v${res.responderVersion})'`);
+  }
+
+  public askForHeartbeats(peers: types.HeardbeatClient[], interval: number = 5000) {
+    setInterval(() => {
+      peers.forEach(async (peer) => {
```
netoneko: We don't really need to wait here, I assume `peers.map(this.askForHearbeat)` would be enough.

PR #76, opened 2018-02-20
Add sync between nodes

netoneko:
1. Node1 asks other nodes if they have blocks newer than X
2. Node1 receives notice from Node2 that it has new blocks
3. Node1 marks Node2 as the one it syncs from and stops accepting `addBlock` requests
4. Node1 asks Node2 to send new blocks
5. Node2 sends new blocks.
6. Node1 collects new blocks
7. Node1 adds blocks to the storage
8. Node1 asks other nodes if they have blocks newer than Y
9. Node1 receives a notification from Node2 that it does not have new blocks
10. Node1 assumes that it's in sync and starts accepting requests

I think there is some complication in case Node2 goes down forever and than Node1 can't get out of sync cycle, but we can solve it later.


projects/services/storage-service-typescript/package.json
```diff
@@ -15,17 +15,27 @@
...
     "@types/lodash": "^4.14.99",
+    "@types/mocha": "^2.2.47",
     "@types/node": "^9.4.4",
     "@types/supertest": "^2.0.4",
+    "chai": "^4.1.2",
+    "chai-as-promised": "^7.1.1",
     "concurrently": "^3.5.1",
+    "fs-extra": "^2.0.0",
+    "mocha": "^5.0.1",
+    "ts-mocha": "^1.1.0",
```
lbeder: Please remove ts-mocha

projects/libs/core-library-typescript/src/common-library/grpc.ts
```diff
@@ -25,13 +24,18 @@ export interface GRPServers {
   endpoint: string;
 }
 
+export interface GRPCRuntime {
+  app: Mali;
+  services: GRPCRuntime[];
```
lbeder: Service

projects/libs/core-library-typescript/src/common-library/grpc.ts
```diff
@@ -40,10 +44,12 @@ function runServer(grpc: GRPServer) {
 
   app.use(serviceFuncs);
   app.start(grpc.endpoint);
+
+  return {app, services: [grpc.service] };
```
lbeder: `{ `

PR #78, opened 2018-02-20
Removed topology out of services

electricmonk:


projects/architecture/interfaces/block-storage.proto
```diff
@@ -1,10 +1,8 @@
 syntax = "proto3";
 
-import "primitives/heartbeat.proto";
```
lbeder: These are all changes that were already merged to `master`. Did you rebase properly?

projects/libs/core-library-typescript/src/base-service/service-runner.ts
```diff
@@ -1,27 +1,20 @@
 import { Service } from "./service";
 
 export class ServiceRunner {
-  public static async run(grpcServerFunc: any, service: Service): Promise<void> {
+  public static async run(grpcServerFunc: any, service: Service, endpoint?: string): Promise<void> {
     await service.start();
 
     return grpcServerFunc({
-      endpoint: service.nodeTopology.endpoint,
+      endpoint: endpoint || service.nodeTopology.endpoint,
```
lbeder: @electricmonk why should `endpoint` be optional and otherwise taken from the `service.nodeTopology.endpoint` singleton? Why shouldn't we just require the `endpoint` parameter and avoid accessing the topology here whatsoever?

projects/libs/core-library-typescript/src/base-service/service-runner.ts
```diff
@@ -1,27 +1,20 @@
...
-      const serviceEndpoint = service.nodeTopology.endpoint;
-      if (endpoint !== serviceEndpoint) {
-        throw new Error(`Multi-services should run on the same endpoint! Expected: ${endpoint}, received: ${serviceEndpoint}`);
-      }
-
       await service.start();
     }
 
-    return grpcServerFunc({ endpoint, services });
+    return grpcServerFunc({ endpoint: endpoint || services[0].nodeTopology.endpoint, services });
```
lbeder: ^ Same here.

projects/libs/core-library-typescript/src/base-service/service.ts
```diff
@@ -34,39 +38,23 @@ export abstract class Service {
     return Service.RPCMethod(target, propertyKey, descriptor, true);
   }
 
-  public constructor() {
+  public constructor(serviceConfig?: ServiceConfig) {
     this.name = this.constructor.name;
     this.nodeTopology = topology();
     this.peers = topologyPeers(this.nodeTopology.peers);
+    this.nodeName = (serviceConfig && serviceConfig.nodeName) || this.nodeTopology.name;
```
lbeder: I think it'd be cleaner to just require the `serviceConfig` parameter here as well.

projects/services/consensus-service-typescript/src/consensus-service.ts
```diff
@@ -15,21 +15,21 @@ export default class ConsensusService extends Service {
   private virtualMachine: types.VirtualMachineClient;
   private blockStorage: types.BlockStorageClient;
 
-  public constructor() {
-    super();
+  public constructor(gossip: types.GossipClient, virtualMachine: types.VirtualMachineClient, blockStorage: types.BlockStorageClient, serviceConfig: ServiceConfig) {
+    super(serviceConfig);
+    this.gossip = gossip;
```
lbeder: Just a side note: you can use a member assignment shortcut. 

Instead of:

```ts
private gossip: types.GossipClient;

public constructor(gossip: types.GossipClient ...
```

You can achieve the same via:

```ts
// You don't need me now! private gossip: types.GossipClient;

public constructor(private gossip: types.GossipClient ...
```

projects/libs/core-library-typescript/src/sidechain-connector/sidechain-connector.ts
```diff
@@ -28,7 +29,7 @@ export class SidechainConnector {
     return EthereumConnector.createHttpConnector(address);
   }
 
-  constructor(options: SidechainConnectorOptions = {}) {
+  constructor(options: SidechainConnectorOptions = { nodeName: "unnamed" }) {
```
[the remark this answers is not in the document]
electricmonk: I don't think this config should be optional at all - I just wasn't sure why it's optional so I didn't want to presume

projects/architecture/interfaces/block-storage.proto
electricmonk: it's the same PR, dunno why this change appears again

projects/libs/core-library-typescript/src/base-service/service-runner.ts
electricmonk: you're right - this is a middle step I did during refactoring to support both approaches, I'll be pushing another commit which removes them

electricmonk:
Re: `NODE_NAME` - I want to, but it seems greedy for the purposes of this refactor. Someone else should probably do that opportunistically when in the general vicinity.

PR #81, opened 2018-02-28
Remove config singleton and make sure that services can be externally configured in full

lbeder:


projects/services/consensus-service-typescript/src/consensus-service.ts
```diff
@@ -31,7 +30,7 @@ export default class ConsensusService extends Service {
 
   @Service.SilentRPCMethod
   public async gossipMessageReceived(rpc: types.GossipMessageReceivedContext) {
-    const obj: any = JSON.parse(rpc.req.Buffer.toString("utf8"));
-    this.consensus.gossipMessageReceived(rpc.req.FromAddress, rpc.req.MessageType, obj);
+    const obj: any = JSON.parse(rpc.req.buffer.toString("utf8"));
```
netoneko: I suggest we start using `payload` instead of `obj`.

projects/services/storage-service-typescript/src/index.ts
```diff
@@ -1,24 +1,42 @@
...
-  nodeName: nodeTopology.name,
-  pollInterval: Number(BLOCK_STORAGE_POLL_INTERVAL) || 5000
-};
 
-const main = async () => {
-  await ServiceRunner.runMulti(grpc.storageServiceServer, [
-    new BlockStorageService(peers.gossip, nodeConfig),
-    new StateStorageService(peers.blockStorage, nodeConfig)
-  ], nodeTopology.endpoint);
+const blockStorageDBPath = path.resolve(path.join("../../../db", NODE_ENV || "development") + "blocks.db");
```
netoneko: Maybe we could use `BLOCK_STORAGE_DB_PATH` instead of constructing a relative path?

projects/services/consensus-service-typescript/src/consensus-service.ts
lbeder: Sure thing

projects/services/storage-service-typescript/src/index.ts
lbeder: Do you mean to take from `process.env` (and if it doesn't exist - to use this default value)?

netoneko: Yes.

lbeder: You got it 👍

projects/services/consensus-service-typescript/src/consensus-server.ts
```diff
@@ -1,21 +1,38 @@
...
-  }
+const DEFAULT_CONSENSUS_CONFIG = {
+  electionTimeout: {
+    min: 2000,
+    max: 4000
+  },
+  heartbeatInterval: 100,
+};
+
+const { NODE_NAME, NUM_OF_NODES, ETHEREUM_CONTRACT_ADDRESS } = process.env;
```
idandash: We better switch to nconf and let it parse the config params automatically. Also, this way, we can conveniently pass them as command line arguments if needed

projects/services/storage-service-typescript/src/index.ts
```diff
@@ -1,24 +1,42 @@
-import { ErrorHandler, grpc, ServiceRunner, topology, topologyPeers } from "orbs-core-library";
+import * as path from "path";
+
+import { logger, ErrorHandler, grpc, ServiceRunner, topology, topologyPeers } from "orbs-core-library";
 
 import BlockStorageService from "./block-storage-service";
 import StateStorageService from "./state-storage-service";
 
+const { NODE_NAME, NODE_ENV, LOGZIO_API_KEY, BLOCK_STORAGE_POLL_INTERVAL, BLOCK_STORAGE_DB_PATH } = process.env;
```
idandash: There's a lot of redundancy here.  It makes much more sense to introduce a Logger.loadConfigFromEnv() or something of that sort

PR #83, opened 2018-02-28
Switch to block-based consensus

idandash:
The flow is as follows:
1. Every new transaction is added to the pool and broadcast to all other nodes. De-dup is currently prevented using a simple hash on the transaction
2. When a new leader is elected and every time a block is committed, the leader switches to a "block append ready" state which makes it attempt a generation of a new block and append it to all nodes.
3. When a block is committed, all its transactions are cleared from the pool.


projects/libs/core-library-typescript/test/virtual-machine/virtual-machine.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+            aTransaction({from: "account1", to: "account3", amount: 2})  // account1 = 1, account2 = 7, account3 = 2
+        ]});
+        stateDiff.should.have.lengthOf(3);
+        for (const item of stateDiff) {
+            item.should.have.property("contractAddress", "foobar");
+        }
+        stateDiff.find(item => item.key === "balances.account1").should.have.property("value", "1");
+        stateDiff.find(item => item.key === "balances.account2").should.have.property("value", "7");
+        stateDiff.find(item => item.key === "balances.account3").should.have.property("value", "2");
+    });
```
netoneko: Can you please add couple of test cases:

1. two identical transactions in one block
2. block with wrong order of transactions
3. unhappy flow

projects/libs/core-library-typescript/src/consensus/raft-consensus.ts
```diff
@@ -103,61 +112,77 @@ export class RaftConsensus {
...
+    if (this.node.isLeader()) {
+      this.readyForBlockAppend = true;
+      logger.info(`Node ${this.node.id} was elected as a new leader!`);
+    } else {
+      this.readyForBlockAppend = false;
+    }
+  }
+
+  private pollForPendingTransactions() {
+    setInterval(async () => {
```
netoneko: Please make the interval a property of RaftConsenus and clear it explicitly in the `async shutdown`.

projects/libs/core-library-typescript/src/consensus/raft-consensus.ts
```diff
@@ -103,61 +112,77 @@ export class RaftConsensus {
...
+      try {
+        if (this.readyForBlockAppend) {
+          logger.debug(`attempting append of new block`);
```
netoneko: Please log the block id or hash so we can identify faulty blocks later.

projects/libs/core-library-typescript/src/consensus/raft-consensus.ts
```diff
@@ -103,61 +112,77 @@ export class RaftConsensus {
...
+          await this.appendNextBlock();
+        }
+      } catch (err) {
+        logger.error("newBlockAppendTick error: " + err);
       }
-    });
+    }, 500);
```
netoneko: Please make it configurable.

projects/libs/core-library-typescript/src/consensus/block-builder.ts
```diff
@@ -0,0 +1,36 @@
+import { types } from "../common-library";
+
+export default class BlockBuilder {
+    private virtualMachine: types.VirtualMachineClient;
+    private transactionPool: types.TransactionPoolClient;
+
+    constructor(input: {virtualMachine: types.VirtualMachineClient, transactionPool: types.TransactionPoolClient}) {
```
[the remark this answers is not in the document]
idandash: Ok. I find it more readable on the caller's side this way

projects/libs/core-library-typescript/src/transaction-pool/transaction-pool.ts
```diff
@@ -1,23 +1,66 @@
...
-    //   logger.info(`${topology.name}: after adding we have ${this.pendingTransactions.size} pending transactions`);
-    // }
+    await this.storePendingTransaction(transaction);
+    await this.broadcastTransactionToOtherNodes(transaction);
+  }
+
+  public getAllPendingTransactions(): types.GetAllPendingTransactionsOutput {
+    // TODO: pull FIFO
+    // TODO: solve concurrency issues
+    const transactions = Array.from(this.pendingTransactions.values());
```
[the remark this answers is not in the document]
idandash: You're right. I added a TODO comment on that myself.

projects/libs/core-library-typescript/src/transaction-pool/transaction-pool.ts
```diff
@@ -1,23 +1,66 @@
...
+    return { transactions };
+  }
+
+  public clearPendingTransactions(transactions: types.Transaction[]) {
```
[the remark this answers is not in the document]
idandash: It's a good point. I was planning to do that later as part of the "transaction appendix mess" we discussed offline

projects/libs/core-library-typescript/src/virtual-machine/virtual-machine.ts
```diff
@@ -15,12 +12,30 @@ export class VirtualMachine {
...
+    for (const transaction of input.orderedTransactions) {
+      try {
+        await this.processor.processTransaction({
+          sender: transaction.sender,
+          contractAddress: transaction.contractAddress,
+          payload: transaction.payload
+        }, stateCache);
+
+        processedTransactions.push(transaction);
+      } catch (err) {
```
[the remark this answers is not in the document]
idandash: Good catch! Nothing happens to the state directly, but I do need to make sure that transactions affect the state cache only if complete successfully. I'll fix it

projects/libs/core-library-typescript/test/consensus/block-builder.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+
+const transactionPool = stubInterface<types.TransactionPoolClient>();
+const virtualMachine = stubInterface<types.VirtualMachineClient>();
+
+function randomUInt() {
+  return Math.floor(Math.random() * Number.MAX_SAFE_INTEGER);
+}
+
+function aRandomTransaction(): types.Transaction {
+    return {
```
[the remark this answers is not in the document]
idandash: There's another "de-dup" task in the backlog that will cover this case

projects/libs/core-library-typescript/src/consensus/raft-consensus.ts
idandash: The shutdown property has not yet been merged to "master". I will do it once it's there

e2e/config/docker/docker-compose.test.services.yml
```diff
@@ -6,7 +6,7 @@ services:
       file: docker-compose.test.base.yml
       service: base-service
     working_dir: /opt/orbs/projects/services/consensus-service-typescript/
-    command: node dist/index.js ${NODE_CONFIG_PATH}/consensus.js
+    command: node --inspect=0.0.0.0:9229 dist/index.js ${NODE_CONFIG_PATH}/consensus.js
```
lbeder: @idandash could you also share the vscode launch settings fikle?

projects/architecture/interfaces/primitives/blocks.proto
```diff
@@ -1,15 +1,21 @@
 syntax = "proto3";
 
-import "./transactions.proto";
+import "transactions.proto";
 
 message BlockHeader {
-  uint32 version = 1;
-  uint32 id = 2;
-  uint32 prevBlockId = 3;
+    uint32 version = 1;
```
lbeder: You've accidentally shifted the indent from 2 to 4 (it's usually 2, see here: https://developers.google.com/protocol-buffers/docs/style)

projects/architecture/interfaces/primitives/blocks.proto
```diff
@@ -1,15 +1,21 @@
...
+    uint32 id = 2;
+    uint32 prevBlockId = 3;
 }
 
+message modifiedStateKey {
```
lbeder: `ModifiedStateKey`

projects/architecture/interfaces/primitives/blocks.proto
```diff
@@ -1,15 +1,21 @@
...
+    string key = 2;
+    string value = 3;
+}
+  
 message Block {
-  BlockHeader header = 1;
-  Transaction tx = 2;
-  string modifiedAddressesJson = 3;
+    BlockHeader header = 1;
+    repeated Transaction transactions = 2;
```
lbeder: 👍 👍 👍

projects/libs/core-library-typescript/src/consensus/block-builder.ts
```diff
@@ -0,0 +1,36 @@
+import { types } from "../common-library";
+
+export default class BlockBuilder {
+    private virtualMachine: types.VirtualMachineClient;
```
lbeder: You are accidentally using 4 spaces in this file as well. If you'd configure vscode to use `.editorconfig`, it'll fix this automatically for you.

projects/libs/core-library-typescript/src/consensus/block-builder.ts
```diff
@@ -0,0 +1,36 @@
...
+    constructor(input: {virtualMachine: types.VirtualMachineClient, transactionPool: types.TransactionPoolClient}) {
+        this.virtualMachine = input.virtualMachine;
+        this.transactionPool = input.transactionPool;
+    }
+
+    public async buildNextBlock(lastBlockId: number) {
+        const { transactions } = await this.transactionPool.getAllPendingTransactions({});
+
+        if (transactions.length == 0) {
+            throw "transaction pool is empty";
```
lbeder: Could you use here and below `throw new Error("transaction pool is empty");`) instead?

projects/libs/core-library-typescript/src/consensus/raft-consensus.ts
```diff
@@ -4,8 +4,10 @@ import { EventEmitter } from "events";
 import { logger } from "../common-library/logger";
 import { types } from "../common-library/types";
 import { config } from "../common-library/config";
+import BlockBuilder from "./block-builder";
 
 import { Gossip } from "../gossip";
+import { Block } from "web3/types";
```
lbeder: Why do we need a definition from `web3` here? Is it by mistake?

projects/libs/core-library-typescript/src/transaction-pool/transaction-pool.ts
```diff
@@ -1,23 +1,66 @@
...
+      const txHash = this.calculateTransactionHash(transaction);
+      this.pendingTransactions.delete(txHash);
+    }
+  }
+
+  public async gossipMessageReceived(fromAddress: string, messageType: string, message: types.GossipMessageReceivedData) {
+    if (messageType == "newTransaction") {
+      this.storePendingTransaction(message.transaction);
+    } else {
+      throw `Unsupported message type ${messageType}`;
```
lbeder: `throw new Error`

projects/libs/core-library-typescript/src/transaction-pool/transaction-pool.ts
```diff
@@ -1,23 +1,66 @@
...
+  }
+
+  private calculateTransactionHash(transaction: types.Transaction) {
+    const hash = createHash("sha256");
+    hash.update(JSON.stringify(transaction));
+    return hash.digest("hex");
+  }
+
+
+
```
lbeder: Some extra blank lines

projects/libs/core-library-typescript/test/consensus/block-builder.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+    transactions.push(aRandomTransaction);
+  }
+
+  return transactions;
+}
+
+function aRandomStateDiff(): types.modifiedStateKey[] {
+  return [
+    {
+      contractAddress: randomUInt().toString(16),
```
lbeder: Generating random numbers in tests is considered a bad practice, as long as there is no way to use a seed and artifact it, so that it'd be possible to replay the tests again and again (in order to reproduce/fix the issue). This is also true when randomizing test order, timeouts/delays, etc.

In this case, could you just generate once a pool of numbers and test it as a table test? Alternatively, we can add having a `process.env.TEST_SEED` and make sure that it's exported or logged, but I personally think that the former would be much simpler and cleaner at this point.

projects/libs/core-library-typescript/test/consensus/block-builder.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+virtualMachine.processTransactionSet.returns(processTransactionSetOutput);
+
+const blockBuilder = new BlockBuilder({ virtualMachine, transactionPool });
+
+describe("a block", () => {
+  it("is built with the transactions on transaction pool", async () => {
+    const block = await blockBuilder.buildNextBlock(1);
+
+    block.transactions.should.eql(dummyTransactionSet);
+
```
lbeder: Empty line

projects/libs/core-library-typescript/test/consensus/block-builder.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+};
+virtualMachine.processTransactionSet.returns(processTransactionSetOutput);
+
+const blockBuilder = new BlockBuilder({ virtualMachine, transactionPool });
+
+describe("a block", () => {
+  it("is built with the transactions on transaction pool", async () => {
+    const block = await blockBuilder.buildNextBlock(1);
+
+    block.transactions.should.eql(dummyTransactionSet);
```
lbeder: I've started this discussion with @netoneko as well: we should use the mode modern `expect` instead of `should`, since the former supports many additional capabilities and has a larger community of tools and contributors. In this case, it won't matter that much, but if we won't choose one - we'd find ourselves duplicating testing tools and primitives all over.

projects/libs/core-library-typescript/test/transaction-pool/transaction-pool.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+        payload: "payload"
+    };
+
+    return transaction;
+}
+
+describe("new broadcast transaction", () => {
+  it("is added to the transaction pool", async () => {
+      const tx = aTransaction();
+      transactionPool.addNewPendingTransaction(tx);
```
lbeder: Isn't this method `async` as well?

projects/libs/core-library-typescript/test/virtual-machine/virtual-machine.spec.ts
lbeder: Agreed. In the worst case, at least let's document these tests as `xit` and make sure to add them next time (of course, if you have time, please add them to this PR as well)

projects/libs/core-library-typescript/test/consensus/block-builder.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+const transactionPool = stubInterface<types.TransactionPoolClient>();
+const virtualMachine = stubInterface<types.VirtualMachineClient>();
+
+function randomUInt() {
+  return Math.floor(Math.random() * Number.MAX_SAFE_INTEGER);
+}
+
+function aRandomTransaction(): types.Transaction {
+    return {
+        version: 1,
```
lbeder: We are currently using `version = 0` for transactions (and appendices) and have a test in `TransactionHandler` (`if (transaction.version !== 0) {`) and another place - if this test passes, does it mean that we don't verify transaction structure at this endpoint? Otherwise, any test using `version = 1` should've failed because of it.

projects/libs/core-library-typescript/src/consensus/raft-consensus.ts
idandash: There's no specific block to log at this stage

projects/libs/core-library-typescript/test/transaction-pool/transaction-pool.spec.ts
idandash: It is. I'll fix

projects/libs/core-library-typescript/test/consensus/block-builder.spec.ts
idandash: 👍

projects/libs/core-library-typescript/test/virtual-machine/virtual-machine.spec.ts
idandash: #2 and #3 are already covered.

idandash: #1 - I'm adding an xit(). I'll implement it as part of the de-dup

PR #87, opened 2018-03-08
transaction/block format restructure

idandash:


projects/architecture/interfaces/primitives/addresses.proto
```diff
@@ -0,0 +1,8 @@
+syntax = "proto3";
+
+message UniversalAddress {
+  uint32 scheme = 1;
+  uint32 networkId = 2;
+  bytes id = 3;
+  uint32 checksum = 4;
+}
```
lbeder: 1. Why do we need so many bytes for `scheme` or `networkId`? 
2. In general, we must to have a way to serialize/deserialize these messages in a fully binary compatible way (e.g., so that we won’t different hashes or signatures). 
3. Why is the checksum a `uint32`? In other words, are we confident that 4 bytes (e.g., 4 bytes of the hash of the address) are enough?

projects/architecture/interfaces/primitives/transactions.proto
```diff
@@ -1,15 +1,32 @@
...
 }
 
-message TransactionAppendix {
-  uint32 version = 1;
-  repeated string prefetchAddresses = 2;
-  string subscriptionKey = 3;
+message TranasactionLifespan {
+  bytes refBlockHash = 1;
+  uint32 blocksToLive = 2;
+}
```
lbeder: Should be assume single hashing algorithm for all messages (e.g., `blockHash`) or would want to support different hashes?

projects/architecture/interfaces/primitives/transactions.proto
```diff
@@ -1,15 +1,32 @@
...
+
+message TransactionBody {
+  bytes contractAddress = 1;
+  bytes payload = 2;
+}
+
+message Transaction {
+  TransactionHeader header = 1;
+  TransactionBody body = 2;
+  bytes signature = 3;
```
lbeder: Can we assume single signature scheme (e.g., that the algorithm will be derived from the address)?

.vscode/launch.json
```diff
@@ -0,0 +1,80 @@
+{
```
lbeder: Superb! 🏆

e2e/src/orbs-client.ts
```diff
@@ -57,25 +55,29 @@ export class OrbsClientSession {
   }
 
   async call(contractAddress: string, payload: string) {
-    const { resultJson } = await this.orbsClient.call({
+    const { resultJson } = await this.orbsClient.callContract({
       sender: this.getAddress(),
-      contractAddress: contractAddress,
+      contractAddress: {address: contractAddress},
```
lbeder: Why do we need another level of indirection here? I.e., why not `contractAddress,`?

e2e/src/orbs-client.ts
```diff
@@ -57,25 +55,29 @@ export class OrbsClientSession {
...
     return {
-      version: 0,
-      sender: this.senderAddress,
-      contractAddress: contractAddress,
-      payload: payload,
-      signature: "" // TODO: add a signature once implement in the network-side
+      header: {
+        version: 0,
+        sender: this.getAddress(),
+        sequenceNumber: 0
```
lbeder: The `sequence` here is confusing, compared how it's used in other blockchains (e.g., Ethereum, Stellar, Ripple, etc.). Maybe something with `order` instead?

projects/architecture/interfaces/primitives/addresses.proto
```diff
@@ -0,0 +1,12 @@
+syntax = "proto3";
+
+message UniversalAddress {
```
lbeder: This will change, according to the last discussions with architecture. I'll address its affects on the SDK and UniversalAddress in a future PR.

projects/architecture/interfaces/primitives/addresses.proto
```diff
@@ -0,0 +1,12 @@
...
+  uint32 scheme = 1;
+  uint32 networkId = 2;
+  bytes id = 3; // TODO: is id the right name? maybe key?
```
lbeder: Today, we refer to it as `accountId`.

projects/architecture/interfaces/primitives/addresses.proto
```diff
@@ -0,0 +1,12 @@
...
+  uint32 checksum = 4;
+}
+
+message ContractAddress {
+  string address = 1; // TODO: what's the format?
```
lbeder: IMHO, it'll be a `UniversalAddress` as well, but now it's not the time anyway.

projects/architecture/interfaces/primitives/blocks.proto
```diff
@@ -1,21 +1,27 @@
...
 message BlockHeader {
   uint32 version = 1;
-  uint32 id = 2;
-  uint32 prevBlockId = 3;
+  bytes prevBlockHash = 2;
+  uint32 height = 3;
+}
+
+message BlockBody {
+  repeated Transaction transactions = 2;
```
lbeder: Let's start from 1, while we still can?

projects/architecture/interfaces/primitives/blocks.proto
```diff
@@ -1,21 +1,27 @@
...
   string key = 2;
   string value = 3;
 }
 
 message Block {
   BlockHeader header = 1;
-  repeated Transaction transactions = 2;
-  repeated ModifiedStateKey stateDiff = 3;
+  BlockBody body = 2;
+  bytes hash = 3; // TODO: how is the hash calculated?
```
lbeder: Do we need a hash here at all? It can be quickly derived from everything (e.g., there is no block hash in either Bitcoin or Ethereum).

projects/architecture/interfaces/primitives/transactions.proto
```diff
@@ -1,15 +1,36 @@
...
+  // bytes signature = 3;
 }
 
-message TransactionAppendix {
+message TransactionHeader {
   uint32 version = 1;
-  repeated string prefetchAddresses = 2;
-  string subscriptionKey = 3;
+  UniversalAddress sender = 2;
+  // TransactionLifespan lifespan = 3;
```
lbeder: Why have you decided to not address lifespan/ttl here? If it's going to be part of a different PR, could you remove this commented out code from here for now?

projects/libs/core-library-typescript/src/block-storage/block-storage.ts
```diff
@@ -59,56 +65,52 @@ export class BlockStorage {
...
-      throw new Error(`Invalid block ID of block: ${JSON.stringify(block)}!`);
+    if (block.header.height !== this.lastBlock.header.height + 1) {
+      throw new Error(`Invalid block height of block: ${JSON.stringify(block)}!`);
     }
 
-    if (block.header.prevBlockId !== this.lastBlock.header.id) {
-      throw new Error(`Invalid prev block ID of block: ${JSON.stringify(block)}! Should have been ${this.lastBlock.header.id}`);
+    if (!block.header.prevBlockHash.equals(this.lastBlock.hash)) {
+      throw new Error(`Invalid prev block hash of block: ${JSON.stringify(block)}! Should have been ${JSON.stringify(this.lastBlock.hash)}`);
     }
```
lbeder: I don't think we should have a hash in the block at all, but if we will - its verifisation should be here as well.

projects/libs/core-library-typescript/src/block-storage/block-storage.ts
```diff
@@ -59,56 +65,52 @@ export class BlockStorage {
...
   }
 
-  public async getBlock(id: number): Promise<types.Block> {
-    return JSON.parse(await this.db.get<string>(id.toString()));
+  public async getBlock(height: number): Promise<types.Block> {
+    return JsonBuffer.parseJsonWithBuffers(await this.db.get<string>(height.toString()));
```
lbeder: 👍

projects/libs/core-library-typescript/src/common-library/block-utils.ts
```diff
@@ -0,0 +1,38 @@
+
```
lbeder: Extra line

projects/libs/core-library-typescript/src/common-library/block-utils.ts
```diff
@@ -0,0 +1,38 @@
...
+import { types, logger } from ".";
+import { createHash } from "crypto";
+
+export namespace BlockUtils {
+  export function calculateBlockHash(blockComponents: {header: types.BlockHeader, body: types.BlockBody}): Buffer {
+    const hash = createHash("sha256");
```
lbeder: It's common to do a double-sha256 here (i.e., `hash = sha256(sha256(block))`) as a "better safe than sorry" mitigation against a similar attack on sha1. Alternatively, we can do sha3 (keccak) once instead.

projects/libs/core-library-typescript/src/common-library/block-utils.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    return block;
+  }
+
+  export function buildNextBlock(body: types.BlockBody, prevBlock?: types.Block): types.Block {
+    return buildBlock({
+      header: {
+        version: 0,
+        prevBlockHash: prevBlock ? prevBlock.hash : new Buffer(""),
+        height: prevBlock ? prevBlock.header.height + 1 : 0
```
lbeder: This should error. It shouldn't be possible to build a block with `height` 0 - this will always be the Genesis block, so if `prevBlock` is undefined - this method should error.

projects/libs/core-library-typescript/src/common-library/block-utils.ts
```diff
@@ -0,0 +1,38 @@
+
+import { types, logger } from ".";
+import { createHash } from "crypto";
+
+export namespace BlockUtils {
+  export function calculateBlockHash(blockComponents: {header: types.BlockHeader, body: types.BlockBody}): Buffer {
```
lbeder: Instead of a namespace + flat function, could you implement it as a class with static methods instead? It'd be more similar to what we did in other places (and what we will do in other languages).

projects/libs/core-library-typescript/src/consensus/block-builder.ts
```diff
@@ -1,36 +1,103 @@
...
 
-    public async buildNextBlock(lastBlockId: number) {
-        const { transactions } = await this.transactionPool.getAllPendingTransactions({});
+    const { processedTransactions, stateDiff } = await this.virtualMachine.processTransactionSet({ orderedTransactions: transactions });
 
-        if (transactions.length == 0) {
-            throw new Error("Transaction pool is empty");
-        }
+    if (processedTransactions.length == 0) {
+        throw new Error("Transaction pool is empty");
```
lbeder: @netoneko discovered an issue with this approach: unprocessed transactions (regardless if all or only some) aren't removed from the transaction pool(s). Could you address it as well?

projects/libs/core-library-typescript/src/consensus/block-builder.ts
```diff
@@ -1,36 +1,103 @@
...
+    const block = await this.buildBlockFromPendingTransactions(lastBlock);
+
+    this.onNewBlockBuild(block);
+
+    logger.debug(`appended new block ${JSON.stringify(block)}`);
+
+    this.readyForBlockAppend = false;
+  }
+
+
```
lbeder: Extra line

projects/libs/core-library-typescript/src/consensus/raft-consensus.ts
```diff
@@ -7,6 +7,9 @@ import BlockBuilder from "./block-builder";
 
 import { Gossip } from "../gossip";
 import { Block } from "web3/types";
+import { build } from "protobufjs";
+import { SSL_OP_CRYPTOPRO_TLSEXT_BUG } from "constants";
```
lbeder: What's that?

projects/libs/core-library-typescript/src/state-storage/state-storage.ts
```diff
@@ -5,64 +5,67 @@ import { types } from "../common-library/types";
...
-    return new Promise((resolve, reject) => {
-      if (blockId > this.lastBlockId) {
-        if (timeout < 200) {
-          reject(new Error(`Timeout in attempt to read block state (${blockId} != ${this.lastBlockId})`));
-        } else {
-          setTimeout(() => this.waitForBlockState(timeout - 200), 200);
-        }
+    const { block } = await this.blockStorage.getLastBlock({});
+    while (timeout > 0) {
+        if (block.header.height <= this.lastBlockHeight) {
```
lbeder: Indentation

projects/libs/core-library-typescript/test/block-storage/block-storage-sync.spec.ts
```diff
@@ -4,111 +4,99 @@ import * as chaiAsPromised from "chai-as-promised";
...
-  return {
-    header: {
-      version: 0,
-      id: prevBlockId + 1,
-      prevBlockId: prevBlockId
-    },
-    tx: { version: 0, contractAddress: "0", sender: "", signature: "", payload: "{}" },
-    modifiedAddressesJson: "{}"
-  };
+function generateNextNullBlock(lastBlock: types.Block): types.Block {
```
lbeder: Maybe `EmptyBlock`?

projects/libs/core-library-typescript/test/block-storage/block-storage-sync.spec.ts
```diff
@@ -4,111 +4,99 @@ import * as chaiAsPromised from "chai-as-promised";
...
       blockStorageSync.onReceiveBlock(blocks[3]);
 
-      blockStorageSync.getQueueSize().should.be.eql(5);
-
-      const appendBlocksPromise = blockStorageSync.appendBlocks();
-      try {
-        await appendBlocksPromise;
-      } catch (e) {
-      }
+      blockStorageSync.getQueueSize().should.eql(5);
```
lbeder: Let's try to stick to `expect`? There is no benefit of using both at the same time all around the project :(

projects/libs/core-library-typescript/test/consensus/block-builder.spec.ts
```diff
@@ -1,76 +1,109 @@
-import { types } from "../../src/common-library/types";
+import { types,  } from "../../src/common-library/types";
```
lbeder: Extra comma

e2e/src/orbs-client.ts
idandash: It's not. It was just an abstraction layer I added as an interim step towards a new structure, while still keeping the existing addressing scheme

e2e/src/orbs-client.ts
idandash: Maybe. Let's wait for the final spec and change accordingly

projects/libs/core-library-typescript/src/common-library/block-utils.ts
idandash: It was added for testing purposes

projects/libs/core-library-typescript/src/consensus/raft-consensus.ts
idandash: Oops..

projects/libs/core-library-typescript/test/block-storage/block-storage-sync.spec.ts
idandash: 👍

projects/libs/core-library-typescript/test/block-storage/block-storage-sync.spec.ts
idandash: I do, it's just that this module was already using should() and I didn't want to touch it too much

projects/libs/core-library-typescript/src/common-library/block-utils.ts
idandash: It's more a placeholder until the hashing scheme is finalized

projects/architecture/interfaces/primitives/transactions.proto
idandash: It's a placeholder. It was a reasonable compromise since I do aim to implement it very soon

PR #89, opened 2018-03-13
CryptoSDK: Orbs address scheme (version 0)

lbeder:


lbeder:
@idandash @netoneko this PR implemented the first addressing scheme. It'll going to change a little bit in another PR, with some feedback from OdedW

client/crypto-sdk/lib/address.cpp
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    // Check the version (the first byte of the decoded address).
+    uint8_t version = decoded[0];
+    if (version != Address::VERSION) {
+        return false;
+    }
+
+    // Check the checksum of the version + account ID:
+    vector<uint8_t> checksum(decoded.cend() - Address::CHECKSUM_SIZE, decoded.cend());
```
idandash: Can you change the order of the 2-lines? It's somehow confusing to see the suffix extracted before the rest

client/crypto-sdk/lib/address.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+    vector<uint8_t> decoded;
+    try {
+        decoded = Base58::Decode(address);
+    } catch (const logic_error &e) {
+        return false;
+    }
+
+    // Version (1) + Account ID (20) + Checksum (4)
+    if (decoded.size() != 1 + Address::ACCOUNT_ID_SIZE + Address::CHECKSUM_SIZE) {
```
idandash: 1 == Address::VERSION_FIELD_SIZE?

client/crypto-sdk/lib/address.cpp
```diff
@@ -0,0 +1,109 @@
...
+const vector<uint8_t> &Address::GetAccountId() const {
+    return accountId_;
+}
+
+const vector<uint8_t> &Address::GetChecksum() const {
+    return checksum_;
+}
+
+const string Address::ToString() const {
+    // Concatenate the version, the accound ID, and the checksum together.
```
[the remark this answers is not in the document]
lbeder: Done

client/crypto-sdk/lib/address.cpp
```diff
@@ -0,0 +1,109 @@
...
+    vector<uint8_t> rawAddress;
+    rawAddress.push_back(version_);
+    rawAddress.insert(rawAddress.end(), accountId_.cbegin(), accountId_.cend());
+    rawAddress.insert(rawAddress.end(), checksum_.cbegin(), checksum_.cend());
+
+    // Return the BASE58 encoding of the raw address.
+    return Base58::Encode(rawAddress);
+}
+
+bool Address::IsValid(const string &address) {
```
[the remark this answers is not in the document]
lbeder: That's a good idea, but I think it'd be better to postpone it a bit, once we'd start using this in the next layer of the SDK.

client/crypto-sdk/lib/address.cpp
lbeder: Agreed. Fixed.

client/crypto-sdk/lib/address.cpp
lbeder: Done and done

lbeder:
@idandash I've addressed almost all of your feedback (other than error status for `IsValid` which I want to add in the future, once it's going to be used [e.g., how to expose these codes for the next layer of sdks, etc.]).

lbeder:
P.S.,
I'll add the improvements to the scheme in a different PR.

PR #91, opened 2018-03-13
Disable subscription validation by default

netoneko:
It prevents us from accepting transactions on testnet

### DIFF
--- a/e2e/config/docker/docker-compose.test.services.yml
+++ b/e2e/config/docker/docker-compose.test.services.yml
@@ -29,7 +29,7 @@ services:
       service: base-service
     working_dir: /opt/orbs/projects/services/public-api-service-typescript/
     environment:
-      - validateSubscription=true
+      VALIDATE_SUBSCRIPTION: 'true'
     command: node  --inspect=0.0.0.0:9229 dist/index.js ${NODE_CONFIG_PATH}/public-api.js
     ports:
       - ${PUBLIC_API_HOST_PORT}:51151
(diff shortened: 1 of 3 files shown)


e2e/config/docker/docker-compose.test.services.yml
```diff
@@ -29,7 +29,7 @@ services:
       service: base-service
     working_dir: /opt/orbs/projects/services/public-api-service-typescript/
     environment:
-      - validateSubscription=true
+      VALIDATE_SUBSCRIPTION: 'true'
```
lbeder: AFAIK, the quotes are not necessariy

projects/services/public-api-service-typescript/src/index.ts
```diff
@@ -24,6 +24,10 @@ if (!NODE_NAME) {
 
 const nodeTopology = topology();
 const peers = topologyPeers(nodeTopology.peers);
-const nodeConfig = { nodeName: NODE_NAME };
+const validateSubscription = VALIDATE_SUBSCRIPTION === "true";
```
lbeder: You can also do `!!VALIDATE_SUBSCRIPTION`

e2e/config/docker/docker-compose.test.services.yml
netoneko: I'm pretty sure without quotes it will fail on `docker-compose up`

lbeder: Really? `true` is a valid value in a YAML. Oh well.

PR #96, opened 2018-03-14
Extract typescript-sdk, add demo app

netoneko:
This PR needs an extensive discussion (what do we even want from an SDK), but it's a good way to check testnet capabilities besides e2e.

Granted, it's a manual test via client app, but it might come in handy and has already highlighted some scenarios that were broken.

I don't expect it to be merged any time soon.


netoneko:
I am closing this pr in favor of a new one that I will introduce later.

PR #97, opened 2018-03-15
e2e stress test

netoneko:
This test takes a parameter of `E2E_ACCOUNTS_TOTAL`, creates number of accounts and runs transactions between them.

I wanted to run it all the time, but then I discovered it breaks our staging deployment because of two bugs we have: in block storage which is susceptible to DoS attack, and in transaction pool, which does not clear bad transactions.

Used to test #99.


e2e/src/perpetual-test.ts
```diff
@@ -0,0 +1,78 @@
...
+async function createAccounts(input: { seed: number, numberOfAccounts: number }): Promise<FooBarAccount[]> {
+  return Promise.all(_.range(input.numberOfAccounts).map((num) => {
+    const amountOfBars = num + 10;
+    const senderAddress = `addr_${input.seed}_${amountOfBars}`;
+
+    return aFooBarAccountWith({ senderAddress, amountOfBars });
+  }));
+}
+
+describe("test multiple transactions", async function () {
```
lbeder: `=>`

e2e/src/stress-test.ts
```diff
@@ -0,0 +1,78 @@
...
+
+      return account.transfer({ to: recipient.address, amountOfBars: amount });
+    }));
+
+    await Promise.all(accounts.map(async (account, num) => {
+      console.log(`Account ${account.address} has balance ${await account.getBalance()}`);
+      const isFirst = num === 0;
+      const amount = 10 + (isFirst ? testConfig.stressTest.accounts : num) - 1;
+
+      return account.should.to.have.bars(amount);
```
idandash: Please use expect()

netoneko: Done.

PR #99, opened 2018-03-18
Clear transaction pool

netoneko:
- [x] sync does not interfere with adding new block via grpc call from consensus
- [x] passes stress test of 100 transfers locally
- [x] consistently passes stress test (I didn't run it twice yet)
- [x] clears failed transactions from transaction pool via gossip messages
- [x] more tests for everything
- [x] moved fake gossip and fake ipc to core-library

**REGRESSION**: Through stress test I found out that sometimes we produce different hashes from the blocks, which is quite problematic. I had to turn hash comparison off. We need to address it.


projects/libs/core-library-typescript/src/test-kit/fake-service-ipc-client.ts
```diff
@@ -1,4 +1,4 @@
-export function generateServiceIPCClient<T>(service: any): T {
+export function generateFakeServiceIPCClient<T>(service: any): T {
```
idandash: Why do you call it Fake? It's not Fake. It is a Service IPC Client

projects/libs/core-library-typescript/test/transaction-pool/transaction-pool.spec.ts
```diff
@@ -32,13 +28,53 @@ function aTransaction() {
...
     const tx = aTransaction();
     await transactionPool.addNewPendingTransaction(tx);
+
     const { transactions } = await transactionPool.getAllPendingTransactions();
+
     transactions.should.eql([tx]);
     expect(gossip.broadcastMessage).to.have.been.called;
   });
 
-  xit("two identical transaction are processed only once");
```
idandash: this works only if they are committed to the same block

projects/libs/core-library-typescript/src/test-kit/fake-service-ipc-client.ts
idandash: Why do you call it Fake? It's not Fake. It is a Service IPC Client

projects/libs/core-library-typescript/test/transaction-pool/transaction-pool.spec.ts
```diff
@@ -32,13 +28,53 @@ function aTransaction() {
...
+  it("two identical transaction are processed only once", (done) => {
+    const tx = aTransaction();
+
+    transactionPool.addNewPendingTransaction(tx).then(() => {
```
idandash: Why are you using then() instead of await?

projects/libs/core-library-typescript/test/transaction-pool/transaction-pool.spec.ts
netoneko: True.

projects/libs/core-library-typescript/src/test-kit/fake-service-ipc-client.ts
netoneko: I renamed it back

projects/services/consensus-service-typescript/src/transaction-pool-service.ts
```diff
@@ -8,17 +8,29 @@ import { TransactionPool } from "orbs-core-library";
...
+  private pollInterval: NodeJS.Timer;
+  private pollIntervalMs = 2000;
 
   public constructor(transactionPool: TransactionPool, serviceConfig: ServiceConfig) {
     super(serviceConfig);
     this.transactionPool = transactionPool;
   }
 
   async initialize() {
+    this.pollInterval = setInterval(() => {
```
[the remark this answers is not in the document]
idandash: It's a bit hacky but we can wait with a better solution for the service management layer. For now:
1. Please make it more maintainable and readable by moving this logic to separate functions (startPoolSizeMonitor() or something of that sort).
2. For the sake of memory optimization, you better export a new getPoolSize() method that calculates the size of the pool from the raw queue instead of fetching all transactions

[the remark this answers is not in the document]
netoneko: Done.

projects/services/consensus-service-typescript/src/transaction-pool-service.ts
```diff
@@ -8,17 +8,32 @@ import { TransactionPool } from "orbs-core-library";
...
     super(serviceConfig);
     this.transactionPool = transactionPool;
   }
 
   async initialize() {
+    this.startPoolSizeMonitor();
   }
 
   async shutdown() {
+    clearInterval(this.pollInterval);
```
idandash: @netoneko stopPoolSizeMonitor()?

netoneko: I think we'll be good with this in the mean time.

PR #100, opened 2018-03-19
Enable Subscription Manager on Staging

lbeder:


lbeder:
@netoneko please use the following address on ROPSTEN Ethereum testnet: 0x67121034220a01b004171ae4a5e60992b879863d

PR #103, opened 2018-03-22
Example slack bot

netoneko:
This PR replaces #96 

I suppose that the stuff from `orbs-client.ts` could be moved to `orbs-javascript-sdk` because it's a duplicate of what we have in E2E.


client/examples/slack-bot/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "orbs-typescript-sdk",
+  "version": "0.1.0",
+  "description": "Orbs Typescript SDK",
```
lbeder: `Orbs Magic Money Slack Bot` or something like that.

client/examples/slack-bot/.gitignore
```diff
@@ -0,0 +1,35 @@
+lib-cov
```
lbeder: In general, I think that it'd be better to rename the directory from 'slack-bot' to `magic-money-bot` (or something a bit less prudent), as we might have other slack bots here, etc.

client/examples/slack-bot/src/index.ts
```diff
@@ -0,0 +1,126 @@
...
+
+const {
+  SLACK_VERIFICATION_TOKEN,
+  SLACK_TOKEN,
+  CONVERSATION_ID,
+  ORBS_API_ENDPOINT,
+  TRANSACTION_TIMEOUT
+} = process.env;
+
+const BOT_ADDRESS = "0000";
```
lbeder: What does "0000" mean here? Is this supposed to an Orbs address as well?

client/examples/slack-bot/src/index.ts
```diff
@@ -0,0 +1,126 @@
...
+const PULL_REQUEST_AWARD = 100;
+
+const config = {
+  subscriptionKey: "0x0213e3852b8afeb08929a0f448f2f693b0fc3ebe",
```
lbeder: I don't like that this magic constant appears hard coded in so many places, but that's not the fault of this PR. Let's think how to fix it globally elsewhere.

lbeder:
@netoneko please give me a couple of days to integrate Orbs addresses to the bot (I'll add it as soon as the first PR is merged). I'll also add a "get my address" FTW.

PR #108, opened 2018-03-27
Feature/business system errors

electricmonk:
Now throwing a `ValidationError` type from the contract, which has an `expected` field. Only expected errors will register failed transactions, while unexpected errors will propagate.


projects/libs/core-library-typescript/src/virtual-machine/hard-coded-contracts/registry/foobar-smart-contract.ts
```diff
@@ -5,12 +5,12 @@ export default class FooBarSmartContract extends BaseSmartContract {
 
   public async transfer(recipient: string, amount: number) {
     if (amount === 0) {
-      throw new Error("transaction amount must be > 0");
+      throw this.validationError("transaction amount must be > 0");
```
lbeder: Why not to use `new ValidationError(message)` here directly? Seems a bit more idiomatic.

projects/libs/core-library-typescript/src/virtual-machine/virtual-machine.ts
```diff
@@ -29,9 +29,13 @@ export class VirtualMachine {
         }, transactionScopeStateCache);
 
       } catch (err) {
-        logger.error(`transaction ${JSON.stringify(transaction)} failed. error: ${err}`);
-        rejectedTransactions.push(transaction);
-        continue;
+        if (!err.expected) {
+          throw err;
+        } else {
```
lbeder: You don't have to have the `else` here, since `throw err` exists the context.

projects/libs/core-library-typescript/test/virtual-machine/virtual-machine.spec.ts
```diff
@@ -50,6 +54,23 @@ describe("test virtual machine", () => {
...
   });
 
+  it("rejects a transaction with a non-positive amount", async () => {
+    const transaction = aTransaction({ from: "account1", to: "account2", amount: 0 });
+
+    const { processedTransactions, stateDiff, rejectedTransactions } = await virtualMachine.processTransactionSet({
+      orderedTransactions: [transaction]
+    });
+
+    rejectedTransactions.should.have.lengthOf(1);
```
lbeder: Can you change to `expect`? We're trying to use a single idiom in our tests (alas, without too much luck).

projects/libs/core-library-typescript/src/virtual-machine/hard-coded-contracts/registry/foobar-smart-contract.ts
electricmonk: I want to control how the exception is being constructed, so a factory method would be a good way to contain the logic in one place.
It also keeps the concrete contract class unaware of the actual class representing the error.

projects/libs/core-library-typescript/src/virtual-machine/virtual-machine.ts
electricmonk: I prefer this style. I like code branches in the same depths to have the same indentation.

projects/libs/core-library-typescript/test/virtual-machine/virtual-machine.spec.ts
electricmonk: Well, the test was like this before and I wasn't sure if this was on purpose or not. I'll change.

projects/libs/core-library-typescript/src/virtual-machine/virtual-machine.ts
lbeder: Alrighty.

projects/libs/core-library-typescript/src/virtual-machine/hard-coded-contracts/registry/foobar-smart-contract.ts
lbeder: Ah, I see.

PR #111, opened 2018-03-28
Transaction Dedup

idandash:
Changes:
1. Transaction had a new timestamp field that is used for determining the "time to live" of the transaction, which afterwards is cleaned up from the transaction pool.
2. A new committed transaction pool stores all committed transactions until expire. When a new transaction is attempted to be added to the "pending transaction pool" it will be dropped if another transaction with the same hash is already in one of the pools, either the pending pool or committed pool.
3. Transactions are moved to the committed pool through a call from the block storage once a new block is committed
4. Basic transaction receipts - generated by the virtual machine and stored in the block


projects/architecture/interfaces/primitives/transactions.proto
```diff
@@ -34,3 +27,13 @@ message SubscriptionSignedTransaction {
   Transaction transaction = 1;
   TransactionSubscriptionAppendix transactionSubscriptionAppendix = 2;
 }
+
+message TransactionReceipt {
+  bytes txHash = 1;
+  bool success = 2; // TODO: This is just a temporary field to distinguish between successful and failed transactions
```
netoneko: Why not `string status`?

projects/libs/core-library-typescript/src/consensus/raft-consensus.ts
```diff
@@ -124,14 +124,12 @@ export class RaftConsensus {
     // committed entry.
     const block: types.Block = JsonBuffer.parseJsonWithBuffers(JSON.stringify(msg.block));
 
+    logger.debug(`new block to be committed ${JSON.stringify(block)}`);
```
netoneko: Please start log messages with capital letter.

projects/libs/core-library-typescript/src/common-library/transaction-utils.ts
```diff
@@ -0,0 +1,15 @@
+import { types } from "../common-library/types";
+import { createHash } from "crypto";
+
+export namespace TransactionUtils {
+  export function calculateTransactionHash(transaction: types.Transaction): Buffer {
+    const hash = createHash("sha256");
+    hash.update(JSON.stringify(transaction.header));
```
netoneko: Can we order these objects somehow? Because this hash is non-deterministic and we can't rely on it, all network comes to a halt after a very short time.

netoneko: Maybe use something like this https://github.com/substack/json-stable-stringify or write our own implementation.

projects/libs/core-library-typescript/src/transaction-pool/README.md
```diff
@@ -1 +1,9 @@
 # Transaction Pool
+
+**
+The transaction pool is purposed for:
+1. Storing transactions that are pending inserstion to a block
+2. Preventing transaction de-duplication
+any
+** How Transaction de-dupication works? **
+For every transaction that is added to the pool, a
```
netoneko: The README is a bit incomplete.

projects/libs/core-library-typescript/test/consensus/block-builder.spec.ts
```diff
@@ -160,16 +107,20 @@ describe("a block", () => {
...
 
       setTimeout(() => {
-        expect(newBlockBuildCallback).to.have.been.calledWith(sinon.match.has("body",
-        sinon.match.has("transactions", dummyTransactionSet).and(sinon.match.has("stateDiff", dummyStateDiff))));
-        done();
-        blockBuilder.stop();
+        try {
+          const bodyMatch = sinon.match.has("transactions", dummyTransactionSet)
+          .and(sinon.match.has("stateDiff", dummyStateDiff))
+          .and(sinon.match.has("transactionReceipts"));
```
netoneko: I think maybe there is a need for more extensive tests that validates contents of `transactionReceipts` including transactions that were rejected or failed.

projects/libs/core-library-typescript/src/common-library/transaction-utils.ts
idandash: This hashing mechanism is *very temporary*. Basically we're going to change it as soon as we implement the transaction signature field

projects/architecture/interfaces/primitives/transactions.proto
idandash: This transaction receipt is very basic.  We'll extend it once @OdedWx will introduce the new structure

projects/libs/core-library-typescript/src/consensus/raft-consensus.ts
idandash: +1

projects/libs/core-library-typescript/test/consensus/block-builder.spec.ts
idandash: 👍

idandash: Actually, this should not be in the scope of the Block Builder test. The transaction receipts are propagated from the virtual machine and assumed to be generated correctly

netoneko: Do we have a test like this in virtual machine?

projects/libs/core-library-typescript/src/common-library/transaction-utils.ts
netoneko: I understand it's temporary, but the problem is that some transactions will inevitably fail to clear out of transaction polls on some nodes.

idandash: You're right. This is another issue. I'll make it deterministic

PR #112, opened 2018-03-28
Added the ability to configure (hard-code) smart contract list via environment variable

electricmonk:


e2e/src/test-environment/orbs-nodes.ts
```diff
@@ -8,6 +8,7 @@ import { initPublicApiClient } from "../public-api-client";
 
 const DOCKER_CONFIG_PATH = path.resolve(path.join(__dirname, "../../config/docker"));
 const NODE_CONFIG_PATH = "/opt/orbs/config/topologies/discovery/node1";
+const TEST_SMART_CONTRACTS = [{address: "foobar", filename: "foobar-smart-contract"}, {address: "text-message", filename: "text-message-smart-contract"}];
```
netoneko: Can you please add this to `deploy/bootstrap/.env`?

projects/services/virtual-machine-service-typescript/src/index.ts
```diff
@@ -14,12 +15,23 @@ if (!NODE_NAME) {
   throw new Error("NODE_NAME can't be empty!");
 }
 
+function parseContractList() {
+  try {
+    return JSON.parse(SMART_CONTRACTS_TO_LOAD);
+  } catch (err) {
+    logger.error(`bad contract list from env: ${SMART_CONTRACTS_TO_LOAD} of type ${typeof SMART_CONTRACTS_TO_LOAD}`);
```
netoneko: Please start log messages with capital letter.

deploy/bootstrap/.env
```diff
@@ -2,3 +2,4 @@ LOG_LEVEL=debug
 GOSSIP_PEERS=ws://us-east-1.global.nodes.staging.orbs-test.com:60001,ws://eu-central-1.global.nodes.staging.orbs-test.com:60001,ws://ap-northeast-1.global.nodes.staging.orbs-test.com:60001,ws://ap-northeast-2.global.nodes.staging.orbs-test.com:60001,ws://ap-southeast-2.global.nodes.staging.orbs-test.com:60001,ws://ca-central-1.global.nodes.staging.orbs-test.com:60001
 E2E_PUBLIC_API_ENDPOINT=0.0.0.0:51151
 E2E_NO_DEPLOY=true
+SMART_CONTRACTS_TO_LOAD=[] # [{address: "foobar", filename: "foobar-smart-contract"}]
```
idandash: This should be set only for non-production environments. I didn't understand by this env filename what setup it's used for

netoneko: We use this file on staging as is, so please uncomment it.

idandash: Let's put a comment that this is just a sample of how the .env file can look like, or even comment out the explicit staging settings and make it fail by default (this will force users to explicitly uncomment and set it)

PR #122, opened 2018-04-01
Introducing key management (again)

netoneko:
Four sets of keys:
1. public keys for gossip signature verification
2. public keys for block signature verification
3. private key for signing gossip messages
4. private key for signing a block

Distinction between gossip and block keys is made because every service sends gossip messages, but only one service signs the block. We don't want block keys to be exposed if someone breaks into public api container.

Roadmap:
- [x] pass private keys to CF as password params (not visible in Amazon UI)
- [x] mount private keys inside containers
- [x] pass public keys via S3
- [x] mount public keys inside containers

I also had to change serialization of Gossip messages because it was difficult to maintain, replacing custom binary format with JSON.


projects/libs/core-library-typescript/src/common-library/signatures.ts
```diff
@@ -0,0 +1,65 @@
...
+import * as stringify from "json-stable-stringify";
+import * as fs from "fs";
+import * as _ from "lodash";
+
+export interface SignaturesConfig {
+  publicKeysPath?: string;
+  privateKeyPath?: string;
+}
+
+export class Signatures {
```
idandash: A refactoring proposal:

- KeyStore - stores your keypair + public keys of all other nodes.  exports: getMyPrivateKey(), getMyPublicKey(), getNodePublicKey(nodeId). 
- KeyManager - uses the key store to get private/public keys, exports: signMessage(message), verifyMessage(nodeId, message)

idandash: A more minor suggestion:  KeyManager sounds better maybe?

projects/libs/core-library-typescript/src/common-library/signatures.ts
```diff
@@ -0,0 +1,65 @@
...
+      });
+    }
+  }
+
+  private signObject(object: any, key: string): string {
+    const sign = crypto.createSign(this.SIGNATURE_TYPE);
+    const payload = _.isBuffer(object) ? object : stringify(object);
+    sign.update(payload);
+
+    return sign.sign(key, this.SIGNATURE_FORMAT);
```
idandash: This enforcement of the encoding (it's not FORMAT btw, it's ENCODING) will break as soon as we support block signatures since we're planning to pass it as a Buffer there, without any encoding

deploy/src/multi-account.ts
```diff
@@ -5,26 +5,39 @@ import { readFileSync } from "fs";
...
+  try {
+    const csv = parse(readFileSync(path).toString(), { columns: true })[0];
+
+    const credentials = {
+      accessKeyId: csv["Access key ID"],
+      secretAccessKey: csv["Secret access key"]
+    };
+  }
+  catch (e) {
+    console.error(`WARNING: Could not find credentials, proceeding without them`);
```
lbeder: `console.warn`?

deploy/src/multi-account.ts
```diff
@@ -5,26 +5,39 @@ import { readFileSync } from "fs";
...
 const parse = require("csv-parse/lib/sync");
 
+function parseCredentials(path: string) {
+  try {
+    const csv = parse(readFileSync(path).toString(), { columns: true })[0];
+
+    const credentials = {
+      accessKeyId: csv["Access key ID"],
+      secretAccessKey: csv["Secret access key"]
+    };
```
lbeder: Nice! This actually handles both of the formats, since the column names are the same for both :)

projects/libs/core-library-typescript/src/common-library/signatures.ts
lbeder: +1 on @idandash's suggestion. A `Signatures` module is a bit weird.

projects/libs/core-library-typescript/src/common-library/signatures.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export interface SignaturesConfig {
+  publicKeysPath?: string;
+  privateKeyPath?: string;
+}
+
+export class Signatures {
+  config: SignaturesConfig;
+  SIGNATURE_FORMAT: crypto.HexBase64Latin1Encoding = "base64";
+  SIGNATURE_TYPE = "sha256";
```
lbeder: `HASH_TYPE`

projects/libs/core-library-typescript/src/common-library/signatures.ts
```diff
@@ -0,0 +1,65 @@
...
+import * as _ from "lodash";
+
+export interface SignaturesConfig {
+  publicKeysPath?: string;
+  privateKeyPath?: string;
+}
+
+export class Signatures {
+  config: SignaturesConfig;
+  SIGNATURE_FORMAT: crypto.HexBase64Latin1Encoding = "base64";
```
lbeder: `SIGNATURE_ENCODING`

projects/libs/core-library-typescript/src/common-library/signatures.ts
```diff
@@ -0,0 +1,65 @@
...
+    this.config = signaturesConfig;
+
+    this.readPrivateKey();
+    this.readPublicKeys();
+  }
+
+  private readPrivateKey() {
+    if (this.config.privateKeyPath) {
+      this.privateKey = fs.readFileSync(this.config.privateKeyPath).toString();
+    }
```
lbeder: What happens if `this.config.privateKeyPath` is falsy? Shouldn't it error?

projects/libs/core-library-typescript/src/common-library/signatures.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  private readPublicKeys() {
+    if (this.config.publicKeysPath) {
```
lbeder: What happens if `this.config. publicKeysPath ` is falsy? Shouldn't it error?

projects/libs/core-library-typescript/src/common-library/signatures.ts
```diff
@@ -0,0 +1,65 @@
...
+      fs.readdirSync(this.config.publicKeysPath).forEach((keyName) => {
+        const contents = fs.readFileSync(`${this.config.publicKeysPath}/${keyName}`).toString();
+        this.publicKeys.set(keyName, contents);
+      });
+    }
+  }
+
+  private signObject(object: any, key: string): string {
```
lbeder: `privateKey`

projects/libs/core-library-typescript/src/common-library/signatures.ts
```diff
@@ -0,0 +1,65 @@
...
+    return verify.verify(publicKey, signature, this.SIGNATURE_FORMAT);
+  }
+
+  public sign(object: any): string {
+    return this.signObject(object, this.privateKey);
+  }
+
+  public verify(object: any, signature: string, publicKeyName: string): boolean {
+    const publicKey = this.publicKeys.get(publicKeyName);
+    return this.verifyObject(object, signature, publicKey);
```
lbeder: Can you add a test if such a named key is found?

projects/libs/core-library-typescript/src/gossip/gossip.ts
```diff
@@ -51,37 +52,35 @@ export class Gossip {
...
+      const { sender, hello, recipient, broadcastGroup, objectType, buffer, signature } = rawMessage;
 
       if (this.localAddress == sender) {
         logger.info("Connected to myself. Disconnecting");
         ws.close();
         return;
       }
 
-      if (offset === message.length) {
+      if (hello) {
```
lbeder: image

projects/libs/core-library-typescript/test/common-library/test-private-keys/secret-message-key
```diff
@@ -0,0 +1,51 @@
+-----BEGIN RSA PRIVATE KEY-----
```
lbeder: Let's not include these at all and just generate the test keys in/before `test.sh`.

projects/libs/core-library-typescript/test/gossip/test-private-keys/node0
```diff
@@ -0,0 +1,51 @@
+-----BEGIN RSA PRIVATE KEY-----
```
lbeder: See above. Let's not keep this raw junk in git...

projects/libs/core-library-typescript/src/common-library/signatures.ts
```diff
@@ -0,0 +1,65 @@
...
+    if (this.config.publicKeysPath) {
+      fs.readdirSync(this.config.publicKeysPath).forEach((keyName) => {
+        const contents = fs.readFileSync(`${this.config.publicKeysPath}/${keyName}`).toString();
+        this.publicKeys.set(keyName, contents);
+      });
+    }
+  }
+
+  private signObject(object: any, key: string): string {
+    const sign = crypto.createSign(this.SIGNATURE_TYPE);
```
lbeder: @idandash @netoneko it's hard for me to decide if it's important or not, but we actually shouldn't use RSA, since it's super hard to distribute RSA keys securely (regardless of future threshold encryption or not). The modern approach is to use a DLP (Discrete Logarithm Problem) based encryption (such as ECC) and then distributing key shares is basically a distribution of a scalar. 

Thinking even more about - it's not important for now. We should be able to easily change it in the near future.

projects/libs/core-library-typescript/src/gossip/gossip.ts
netoneko: neil-degrasse-tyson-reveals-the-biggest-misconceptions-people-have-about-the-universe

projects/libs/core-library-typescript/src/common-library/signatures.ts
netoneko: It's an optional parameter. KeyManager is designed in a way that allows it both sign and validate signatures without forcing it to have keys unnecessary for the task.

projects/libs/core-library-typescript/src/common-library/signatures.ts
netoneko: Renamed to `SIGNATURE_ENCODING`. We can update it to output buffers when we need it.

PR #125, opened 2018-04-08
Initial Javascript client SDK

idandash:


lbeder:
As long we don't have package management, I'd limit external packages to as few as possible, so I was thinking of actually adding the above to the existing client SDKs, where some of the functionality is native and some isn't (e.g., not to wrap a javascript wrapper with another javascript wrapper). 

Let's discuss it together before merging it.

netoneko:
@lbeder since js sdk was merged with js-crypto-sdk, I think we can merge this pr now

lbeder:
👍

PR #126, opened 2018-04-09
HTTP interface for public API

netoneko:
Depends on #125

- [x] Introduces http endpoint for public api
- [x] replaces grpc transport in javascript sdk with http transport


projects/libs/core-library-typescript/test/gossip/gossip.spec.ts
```diff
@@ -74,7 +74,7 @@ describe("Gossip", function () {
 
   describe("with signatures", () => {
     before(async function() {
-      this.timeout(6000);
+      this.timeout(20000);
```
electricmonk: huh?

netoneko: Sometimes it takes more times to generate keys and it fails for that reason.

PR #130, opened 2018-04-12
refactor storage service to GRPCerverBuilder and service test + fix for stopping state storage polling

jlevison:


projects/services/storage-service-typescript/test/storage-service.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+      .withService("Gossip", gossipServerStub)
+      .withService("TransactionPool", transactionPoolStub)
+      .onEndpoint(endpoint);
+    server.start();
+
+    blockClient = grpc.blockStorageClient({ endpoint });
+    stateClient = grpc.stateStorageClient({ endpoint });
+  });
+
+  it("should fetch genesis block for an empty database", async () => {
```
idandash: You must not assume the order of the tests. Please use BeforeEach()

projects/services/storage-service-typescript/test/storage-service.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+        {
+          service: "consensus",
+          endpoint: endpoint,
+        },
+      ],
+    };
+
+    const NODE_NAME = "tester";
+    const BLOCK_STORAGE_POLL_INTERVAL = 5000;
+    const BLOCK_STORAGE_DB_PATH = "./test/db/";
```
idandash: The code folder is not necessarily writable. The DB should be created in a temp folder.

projects/libs/core-library-typescript/src/block-storage/block-storage.ts
```diff
@@ -82,8 +82,11 @@ export class BlockStorage {
   public async getBlocks(fromLastBlockHeight: number): Promise<types.Block[]> {
     const blocks: types.Block[] = [];
 
-    for (let i = fromLastBlockHeight; i < this.lastBlock.header.height; ++i) {
-      blocks.push(await this.getBlock(i + 1));
+    // lastBlock is undefined when the service did not initialize the logic/load() did not run yet (race condition most likely)
```
netoneko: It's not a very good idea to return empty blocks if we actually supposed to have the blocks, I think maybe we should either reject the promise or run `.load()` and wait until it's done.

projects/services/storage-service-typescript/src/server.ts
```diff
@@ -0,0 +1,28 @@
...
+        throw new Error("NODE_NAME can't be empty!");
+    }
+
+    const peers = topologyPeers(nodeTopology.peers);
+    // TODO: this is very fragile - refactor to make this less hardcoded
+    const blockStorageDBPath = BLOCK_STORAGE_DB_PATH || path.resolve(path.join("../../../db", NODE_ENV || "development") + "blocks.db");
+    const blockStorageConfig = {
+      nodeName: NODE_NAME,
+      dbPath: blockStorageDBPath,
+      pollInterval: Number(BLOCK_STORAGE_POLL_INTERVAL) || 5000
```
netoneko: It's interesting that block storage poll interval is 5 seconds while we produces a block every 0.5 seconds.

projects/services/storage-service-typescript/test/storage-service.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+    blockClient = grpc.blockStorageClient({ endpoint });
+    stateClient = grpc.stateStorageClient({ endpoint });
+
+    // return the start promise to delay execution of the tests until its resolved (=services started) mocha plays nicely like that
+    return server.start();
+  });
+
+  it("should fetch genesis block for an empty database", async () => {
+    const lastBlock = await blockClient.getLastBlock({});
+    return expect(lastBlock).to.have.property("block")
```
netoneko: We could just match the whole genesis block instead of looking for a specific property.

projects/libs/core-library-typescript/src/state-storage/state-storage.ts
```diff
@@ -13,19 +13,35 @@ export class StateStorage {
...
+    this.engineRunning = false;
+    this.stopPolling();
   }
 
-  public poll() {
-    this.pollInterval = setInterval(() => this.pollBlockStorageCallback(), 200);
+  private startPolling() {
+    if (this.engineRunning) {
+      this.pollInterval = setInterval(() => this.pollBlockStorageCallback(), this.pollIntervalMs);
+      logger.debug(`polling started, interval is ${this.pollIntervalMs}`);
```
netoneko: Please start a sentence with a capital letter in every logger message.

projects/libs/core-library-typescript/src/state-storage/state-storage.ts
```diff
@@ -48,11 +64,26 @@ export class StateStorage {
...
+      logger.warn(`could not get blocks while polling, last height: ${this.lastBlockHeight}, ${err}`);
+    }
 
-    // Assuming an ordered list of blocks.
-    for (const block of blocks) {
-      await this.syncNextBlock(block);
+    if (blocks != undefined) {
+      // until we finish syncing all of them, lets stop the polling
+      try {
+        this.stopPolling();
```
netoneko: I think it would make sense to stop polling and then ask for blocks to avoid a situation when we have a race condition because of two `getBlocks` calls.

projects/services/storage-service-typescript/src/server.ts
```diff
@@ -6,7 +6,7 @@ import BlockStorageService from "./block-storage-service";
 import StateStorageService from "./state-storage-service";
 
 export default function(nodeTopology: any, env: any) {
-    const { NODE_NAME, NODE_ENV, BLOCK_STORAGE_POLL_INTERVAL, BLOCK_STORAGE_DB_PATH } = env;
+    const { NODE_NAME, NODE_ENV, BLOCK_STORAGE_POLL_INTERVAL, BLOCK_STORAGE_DB_PATH, STATE_STORAGE_POLL_INTERVAL } = env;
```
netoneko: Please add a description of this variable to documentation in `deploy/bootstrap/README.md`.

PR #131, opened 2018-04-12
Contract test for http public api

netoneko:
- [x] introduces http mock server in client
- [x] both mock server and real server pass the same test
- [x] client sdk uses mock server instead of request mocks


client/client-sdk-javascript/test/client-sdk.spec.ts
```diff
@@ -17,37 +18,53 @@ const senderAddress = new Address(publicKey, VIRTUAL_CHAIN_ID, Address.TEST_NETW
...
+};
+
 describe("A client calls the connector interface with the correct inputs when", async function () {
   let orbsClient: OrbsClient;
-  const connection = stubInterface<PublicApiConnection>();
+  let httpServer: Server;
 
   beforeEach(async () => {
     orbsClient = new OrbsClient(API_ENDPOINT, senderAddress.toString(), TIMEOUT);
+    httpServer = mockHttpServer(expectedTransaction, expectedContract).listen(8888);
```
electricmonk: please extract `8888` to a const

client/client-sdk-javascript/test/client-sdk.spec.ts
```diff
@@ -17,37 +18,53 @@ const senderAddress = new Address(publicKey, VIRTUAL_CHAIN_ID, Address.TEST_NETW
...
   it("callContract() is called", async () => {
-    nock(API_ENDPOINT)
-      .post("/public/callContract", (res: any) => {
-        expect(res.contractAddress).to.be.eql({ address: "contractAddress" });
-        expect(res.payload).to.be.eql("call-payload");
-        return true;
-      }).reply(200, { result: 12 });
-
-    expect(await orbsClient.call("contractAddress", "call-payload")).to.be.eql(12);
+    expect(await orbsClient.call("contractAddress", "call-payload")).to.be.eql("some-answer");
```
electricmonk: this test isn't symmetrical - how should the reader know why the server is expected to return `"some-answer"`? it's better to make `mockHttpServer` programmable and to set it up in a previous line.
Not urgent for now, but we *will* need it eventually and I think maybe it's better to do it now so that the test is clearer.

projects/services/public-api-service-typescript/test/public-api-service.spec.ts
```diff
@@ -39,46 +41,134 @@ const contractInput: types.CallContractInput = {
...
+}
+
+class FakeTransactionPool extends Service implements types.TransactionPoolServer {
+  @Service.RPCMethod
+  addNewPendingTransaction(rpc: types.AddNewPendingTransactionContext): void {
+    expect(rpc.req).to.be.eql({ transaction });
+  }
+
+  getAllPendingTransactions(rpc: types.GetAllPendingTransactionsContext): void {
+    throw new Error("Method not implemented.");
```
electricmonk: we need to open a discussion about the code smell that is exposed by the need to implement-but-then-not-really-implement these methods

projects/services/public-api-service-typescript/test/public-api-service.spec.ts
```diff
@@ -39,46 +41,134 @@ const contractInput: types.CallContractInput = {
...
+  let httpService: PublicApiHTTPService;
+  let httpEndpoint: string;
+
+  let grpcService: GRPCServerBuilder;
+
+  describe("Read http api", () => {
+    beforeEach(async () => {
+      const httpPort = await getPort();
+      httpEndpoint = `http://127.0.0.1:${httpPort}`;
+      const endpoint = `0.0.0.0:${await getPort()}`;
```
electricmonk: rename as `grpcEndpoint`

projects/services/public-api-service-typescript/test/public-api-service.spec.ts
```diff
@@ -39,46 +41,134 @@ const contractInput: types.CallContractInput = {
...
-    (<sinon.SinonStub>subscriptionManager.getSubscriptionStatus).returns({active: true, expiryTimestamp: Date.now() + 10000000});
-    (<sinon.SinonStub>virtualMachine.callContract).returns({resultJson: "some-answer"});
-    grpcService = new PublicApiService(virtualMachine, transactionPool, {
-      nodeName: "tester"
+  let httpService: PublicApiHTTPService;
+  let httpEndpoint: string;
+
+  let grpcService: GRPCServerBuilder;
+
+  describe("Read http api", () => {
```
electricmonk: should be "Real HTTP API"

netoneko:
@electricmonk I think I’ve addressed all your comments, please take a look

electricmonk:
@netoneko Please make sure to create tickets for the review comments you did not yet address

PR #132, opened 2018-04-14
Changes in transaction structure and addressing scheme

idandash:
Changes:
- Modified server-side to support the new transaction structure & addressing scheme
  - The HTTP Server expects sender and contract addresses as base58, decodes them, and passes them to the transaction pool and from there to other services pass them as byte arrays
  -  Contracts are passed with a sender address encoded also as base58. For the sake of consistency, The test contracts (foobar and text-messages) also expect addresses passed as transaction arguments in base58 format.
- Made minor changes in the API protocol to support the new transaction structure. In addition, transaction is passed in the body as a root json object (rather as a nested object as before)
- Removed tests of soon-to-be-deprecated public api grpc service


client/client-sdk-javascript/test/client-sdk.spec.ts
```diff
@@ -3,51 +3,50 @@ import  * as chai from "chai";
...
+import { Transaction, SendTransactionInput, CallContractInput } from "orbs-interfaces";
+import * as crypto from "crypto";
 
 chai.use(sinonChai);
 
-const publicKey = new ED25519Key().publicKey;
+const senderPublicKey = new ED25519Key().publicKey;
 const VIRTUAL_CHAIN_ID = "640ed3";
-const senderAddress = new Address(publicKey, VIRTUAL_CHAIN_ID, Address.TEST_NETWORK_ID);
+const senderAddress = new Address(senderPublicKey, VIRTUAL_CHAIN_ID, Address.TEST_NETWORK_ID);
```
electricmonk: this duplicates logic from `OrbsContractAdapter` - change this test to use it

client/client-sdk-javascript/src/orbs-contract-adapter.ts
```diff
@@ -1,15 +1,22 @@
 import { OrbsClient } from "./orbs-client";
 import { SendTransactionOutput } from "orbs-interfaces";
+import { Address } from "./address";
+import * as crypto from "crypto";
 
 export type OrbsContractMethodArgs = [string | number] | any[];
 
 export class OrbsContractAdapter {
```
electricmonk: consider renaming this class, as `Adapter` is a mechanical name that does not add any further information. Maybe `OrbsContractClient`? or simply `OrbsContract`?

e2e/src/test-environment/orbs-nodes.ts
```diff
@@ -7,7 +7,12 @@ import { delay } from "bluebird";
 
 const DOCKER_CONFIG_PATH = path.resolve(path.join(__dirname, "../../config/docker"));
 const NODE_CONFIG_PATH = "/opt/orbs/config/topologies/discovery/node1";
-const TEST_SMART_CONTRACTS = [{address: "foobar", filename: "foobar-smart-contract"}, {address: "text-message", filename: "text-message-smart-contract"}];
+// TODO: vchainId should not be hard-coded but rather propagated from the test config!!!
```
electricmonk: this comment should be in line 11, as it pertains to this specific line

projects/libs/core-library-typescript/src/common-library/address.ts
```diff
@@ -0,0 +1,71 @@
...
+    return bs58EncodeRawAddress(rawAddress);
+  }
+}
+
+export function createContractAddress(contractName: string, vchainId = Address.SYSTEM_VCHAINID) {
+  const publicKey = crypto.createHash("sha256").update(contractName).digest();
+  return new Address(publicKey, vchainId);
+}
+
+export function bs58EncodeRawAddress(rawAddress: Buffer) {
```
electricmonk: consider moving both this and the decode functions to an `AddressCodec` class and then importing it instead of these two everywhere

projects/libs/core-library-typescript/test/consensus/block-builder.spec.ts
```diff
@@ -26,32 +28,23 @@ function aGenesisBlock(): types.Block {
...
 
-function aDummyTransaction(addressId): types.Transaction {
-  const senderAddress: types.UniversalAddress = {
-    id: new Buffer(addressId),
-    scheme: 0,
-    checksum: 0,
-    networkId: 0
-  };
-
+function aDummyTransaction(publicKey: Buffer): types.Transaction {
```
electricmonk: isn't this duplicated from the testkit? (`dummy-transaction.ts`)? 
anyway, I'd refactor both to a file `transaction-builders` which also builds the transaction set.

projects/libs/core-library-typescript/test/gossip/gossip.spec.ts
```diff
@@ -46,7 +45,7 @@ describe("Gossip", function () {
         const consensus = consensuses[i];
         if (i == recipientId) {
           expect(consensus.gossipMessageReceived).to.have.been.calledOnce;
-          expect(consensus.gossipMessageReceived.getCall(0).args[0]).to.have.property("buffer").which.equalBytes(buffer);
+          expect((<sinon.SinonSpy>consensus.gossipMessageReceived).getCall(0).args[0]).to.have.property("buffer").which.deep.equal(buffer);
```
electricmonk: not your commit but I'd follow to boy scout rule and extract an assertion plugin `messageBuffer` or similar. Also consider using `sinon-chai` instead of imperatively extracting stuff out of Sinon

PR #133, opened 2018-04-15
Update README

netoneko:
Removed a lot of outdated things, explained env variables that we currently use.


deploy/bootstrap/docker-compose.yml
```diff
@@ -93,7 +93,7 @@ services:
       - .env
       - .env-secrets
     environment:
-      SERVICE_NAME: virtual-machine
+      SERVICE_NAME: sidechain-connector
```
idandash: 👍

PR #136, opened 2018-04-16
Refactor build

netoneko:
- [x] Split client and server build
- [x] Run e2e
- [x] Add docs

A good low-hanging fruit would be to make including android sdk in sdk base image optional. That would cut down couple of minutes of e2e build on CI, although it would have no effect locally.


projects/services/public-api-service-typescript/build.sh
```diff
@@ -7,4 +7,7 @@ yarn install
 
 yarn run build
 
+echo "Copy mock server from client-sdk-javascript to run contract test..."
+cp ../../../client/client-sdk-javascript/test/mock-server.ts test
```
idandash: It would probably make more sense (but not a showstopper) to move mock-server to a `common` folder, or (maybe even better) move to the server-side as a `testkit` and have it exported, hence available, to the client.

.circleci/config.yml
```diff
@@ -5,13 +5,25 @@ jobs:
     steps:
       - checkout
       - run: ./.circleci/setup-rngd.sh
+      - run: ./docker/build-server-base.sh
```
idandash: Please consider encapsulating the `build` and `deploy` scripts already inside the relevant folders (`client`,`e2e`, `projects` (which should be renamed to `server` very soon)

projects/services/public-api-service-typescript/build.sh
netoneko: I agree, it should be extracted later. I think @electricmonk can take it from here because he's working on the contract test a the moment.

.circleci/config.yml
netoneko: It's a concern that we can solve a bit later. I agree with you generally that it should be done.

Dockerfile.server.base
```diff
@@ -0,0 +1,19 @@
...
+
+ARG CI
+ENV CI=$CI
+ENV PROJECT_TYPE="server"
+
+VOLUME [ "/opt/orbs/logs", "/opt/orbs/db" ]
+
+# TODO: move to later stage, use cache for ts compiler & top level software
+
+ADD package.json yarn.lock .yarnrc docker/bootstrap-*.sh docker/install*.sh /opt/orbs/
```
lbeder: You don't need to include `bootstrap-sdk.sh` or `install*.sh` here (the later is unneeded at all)

README.md
```diff
@@ -51,6 +57,13 @@ Pleaser refer to [documentation](e2e/README.md) in `e2e` folder to learn more ab
 
 ## Development
 
+`PROJECT_TYPE` variable is responsible for building a certain part of the system: `server`, `sdk` or `e2e`.
+
+```bash
+export PROJECT_TYPE=server
+./build.sh
```
lbeder: Please create specific scripts for this as well, just like with Docker.

docker/extract-sdk.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+mkdir -p artifacts
+
+export SDK_CONTAINER=$(docker run -d --rm orbs:sdk sleep 60)
+
+export ANDROID_EAR=client/crypto-sdk-android/crypto-sdk/build/outputs/ear
```
lbeder: `aar`...

docker/extract-sdk.sh
```diff
@@ -0,0 +1,15 @@
...
+export ANDROID_JAR=client/crypto-sdk-android/crypto-sdk/build/outputs/jar
+
+export ARTIFACTS=($ANDROID_JAR $ANDROID_EAR)
```
lbeder: `AAR`

docker/extract-sdk.sh
```diff
@@ -0,0 +1,15 @@
...
+
+for i in $ARTIFACTS; do
+    mkdir -p artifacts/$i
+    docker cp $SDK_CONTAINER:/opt/orbs/$i artifacts/$i
+done
```
lbeder: What about other SDKs? What about `client-sdk-javascript`?

projects/services/public-api-service-typescript/build.sh
```diff
@@ -7,4 +7,14 @@ yarn install
 
 yarn run build
 
+echo "Copy mock server from client-sdk-javascript to run contract test..."
+cp ../../../client/client-sdk-javascript/src/orbs-api-interface.ts src
+cp ../../../client/client-sdk-javascript/test/mock-server.ts test
```
lbeder: This and the .gitignore above are too hacky.. can you reuse the code, without duplicating it like this?

docker/extract-sdk.sh
netoneko: I am not sure we even need that considering you said we are not going to export build artifacts any time soon. Let’s sync about it tomorrow whenever you have time.

projects/services/public-api-service-typescript/build.sh
netoneko: If we extract it as another library and then link it to both client and public api, then yes. If you think it’s necessary, I can do it in no time.

PR #137, opened 2018-04-16
Added GetTransactionStatus() to the Transaction Pool

idandash:
Other interesting changes:
1. TransactionUtils is now introduced as a Class wrapping a transaction (renamed to TransactionHelper since I couldn't find an appropriate name).
2. Hacks to support Protobuf Enums on Typescript


projects/libs/core-library-typescript/src/common-library/transaction-helper.ts
```diff
@@ -0,0 +1,24 @@
...
+  }
+
+  public calculateHash(): Buffer {
+    const hash = createHash("sha256");
+    hash.update(stringify(this.header));
+    hash.update(stringify(this.payload));
+    return hash.digest();
+  }
+
+  public calculateTransactionId(): string {
```
lbeder: Raising this again: let's add a semantic type for important types (e.g., `TxID` instead of `string`).

projects/libs/core-library-typescript/src/test-kit/transaction-builders.ts
```diff
@@ -22,7 +22,7 @@ export function aDummyTransactionSet(numberOfTransactions = 3): types.Transactio
   const transactions: types.Transaction[] = [];
   for (let i = 0; i < numberOfTransactions; i++) {
     const transaction = aDummyTransaction({senderPublicKey: createHash("sha256").update(`address${i}`).digest()});
-    const txHash = TransactionUtils.calculateTransactionHash(transaction);
+    const txHash = new TransactionHelper(transaction).calculateHash();
```
lbeder: Why not have an object for `Transaction`?

projects/libs/core-library-typescript/src/transaction-pool/pending-transaction-pool.ts
```diff
@@ -21,6 +20,20 @@ export class PendingTransactionPool extends BaseTransactionPool {
...
+  public getTransactionStatus(txid: string): types.GetTransactionStatusOutput {
+    let receipt: types.TransactionReceipt;
+    let status: types.TransactionStatus = TransactionStatus.NOT_FOUND;
+
+    if (this.pendingTransactions.has(txid)) {
+      status = types.TransactionStatus.PENDING;
+    } else if (this.committedTransactionPool.hasTransactionWithId(txid)) {
+      receipt = this.committedTransactionPool.getTransactionReceiptWithId(txid);
+      status = types.TransactionStatus.COMMITTED;
+    }
```
lbeder: What about historic transactions?

projects/libs/core-library-typescript/src/transaction-pool/committed-transaction-pool.ts
```diff
@@ -2,34 +2,39 @@ import { logger } from "../common-library/logger";
...
-
-      if (this.isExpired(Number(timestamp))) {
-        continue;
-      }
-
-      this.committedTransactions.set(txid, Number(timestamp));
+  public addCommittedTransactions(transactionReceipts: types.TransactionReceipt[]) {
+    const entryTimestamp = Date.now();
+    for (const receipt of transactionReceipts) {
+      const txid = receipt.txHash.toString("hex");
```
lbeder: Let's move this to `TransactionHelper` as well. Otherwise. it's asymmetrical and repetitive.

projects/services/consensus-service-typescript/src/transaction-pool-service.ts
```diff
@@ -67,6 +67,12 @@ export default class TransactionPoolService extends Service {
     logger.debug(`getAllPendingTransactions() . returns ${JSON.stringify(rpc.res)}`);
   }
 
+  @Service.RPCMethod
+  public async getTransactionStatus(rpc: types.GetTransactionStatusContext) {
+    const txid = rpc.req.txid;
+    this.pendingTransactionPool.getTransactionStatus(txid);
```
lbeder: And if it's no longer pending / was committed a year ago?

projects/libs/core-library-typescript/src/test-kit/transaction-builders.ts
idandash: It's confusing since we have a Transaction as a protobuf-derived interface. That was actually how it was called initially

projects/libs/core-library-typescript/src/transaction-pool/pending-transaction-pool.ts
idandash: Not supported for now.

projects/libs/core-library-typescript/src/transaction-pool/committed-transaction-pool.ts
idandash: I agree. I'll fix

projects/services/consensus-service-typescript/src/transaction-pool-service.ts
idandash: No support for that (for now)

projects/libs/core-library-typescript/src/common-library/transaction-helper.ts
idandash: The problem is that as long as we use protobuf, the only way to achieve that, is by declare this type as a message wrapper for string. It creates too much overhead as well as makes any code handling with it really ugly

PR #143, opened 2018-04-25
Add vagrant for vm installation

netoneko:
Per Idan's requirement to support non-cloud installations.


netoneko:
Idan says the requirement doesn't exist, but I still think it's a good idea to have a placeholder in case our clients don't want to use AWS.

PR #146, opened 2018-04-28
GetTransactionStatus http endpoint

idandash:


client/client-sdk-javascript/src/orbs-client.ts
```diff
@@ -44,6 +43,18 @@ export class OrbsClient {
     return body.result;
   }
 
+  async getTransactionStatus(txid: string): Promise<any> {
```
electricmonk: why not `Promise<TXReceipt>`?

client/client-sdk-javascript/src/orbs-contract.ts
```diff
@@ -19,7 +18,7 @@ export class OrbsContract {
     );
   }
 
-  public async sendTransaction(methodName: string, args: OrbsContractMethodArgs): Promise<SendTransactionOutput> {
+  public async sendTransaction(methodName: string, args: OrbsContractMethodArgs) {
```
electricmonk: what? so how do we get the txid?

projects/services/public-api-service-typescript/test/public-api-service.spec.ts
```diff
@@ -17,6 +17,20 @@ const contractAddressBase58 = "T00LUPVrDh4SDHggRBJHpT8hiBb6FEf2rMkGvQPR";
 const transactionTimestamp = "12345678";
 const payload = JSON.stringify({ foo: "bar" });
 
+const transaction: types.Transaction = {
```
electricmonk: why not use a transaction builder? this test should not be intimately familiar with how addresses in a transaction are encoded

client/client-sdk-javascript/src/orbs-contract.ts
idandash: It's fixed in another PR

projects/services/public-api-service-typescript/test/public-api-service.spec.ts
idandash: IMO it should, because what it tests is that an http request (that has a b58-encoded address) is decoded well before propagated to the TX pool

PR #152, opened 2018-05-03
Feature/orbs contributor bootstrap

IdoZilberberg:
First version of an automatic installation script for Orbs contributors. It sets up a MacOS with the software required to run a successful build, then runs the build.

It's presently on the repo's root, alternative location suggestions are welcome!


orbs-contributor-bootstrap.sh
```diff
@@ -0,0 +1,180 @@
...
+# mkdir -p ${DEV_HOME}
+# cd ${DEV_HOME}
+# git clone ${ORBS_NETWORK_REPO}
+# cd ${ORBS_NETWORK_HOME} || { echo "Failed to create ${DEV_HOME} or failed to clone ${ORBS_NETWORK_REPO}. Cannot continue."; exit 1;}
+
+
+DEV_HOME=${HOME}/dev/orbs
+NODE_VER="v9.11.1"
+
+CASK_PACKAGES="google-chrome visual-studio-code iterm2 atom slack java8 docker sourcetree spectacle alfred"
```
netoneko: Why do we need anything but java8 and docker? It seems very excessive.

orbs-contributor-bootstrap.sh
```diff
@@ -0,0 +1,180 @@
...
+PACKAGES="node@8 typescript yarn"
```
netoneko: We use node 9.

PR #158, opened 2018-05-07
HTTP mock server as an independent process

electricmonk:
It is also programmable from the outside via cli arguments


electricmonk:
@jlevison since I don't know when I'll have a few hours for independent packaging for the mock , here's my work so far for merging purposes, so that you're not blocked

netoneko:
Does it mean we will have two mock servers for some time?

netoneko:
@jlevison are you going to incorporate this in your next p-r?

PR #160, opened 2018-05-09
Added signature verification to the Transaction Validator

idandash:


projects/libs/core-library-typescript/src/transaction-pool/transaction-validator.ts
```diff
@@ -1,22 +1,35 @@
...
+  public async validate(tx: types.Transaction): Promise<boolean> {
+    if (tx.header.version !== 0) {
+      throw new Error(`Invalid transaction version: ${tx.header.version}`);
+    }
 
-  public async validate(transaction: types.Transaction): Promise<boolean> {
-    if (transaction.header.version !== 0) {
-      throw new Error(`Invalid transaction version: ${transaction.header.version}`);
+    if ((this.options.verifySignature) && (!this.verifySignature(tx))) {
+      logger.info(`signature of ${JSON.stringify(tx)} is invalid`);
```
jlevison: logs should begin with a capital letter

projects/libs/core-library-typescript/src/transaction-pool/transaction-validator.ts
```diff
@@ -31,19 +44,27 @@ export class TransactionValidator {
...
 
     if (!status.active) {
-      logger.info("subscription ${subscriptionKey} not active");
+      logger.info(`subscription ${subscriptionKey} not active`);
     }
 
     return status.active;
   }
 
+  private verifySignature(tx: types.Transaction) {
```
jlevison: looks okay, but maybe @lbeder should verify the logic here

projects/services/public-api-service-typescript/src/service.ts
```diff
@@ -61,7 +61,11 @@ export default class PublicApiHTTPService extends Service {
             timestamp: _.get(body, "header.timestamp"),
             contractAddress: bs58DecodeRawAddress(_.get(body, "header.contractAddressBase58")),
           },
-          payload: _.get(body, "payload")
+          payload: _.get(body, "payload"),
+          signatureData: { // TEMPORARY PLACEHOLDER
```
jlevison: was this left here intentionally?

projects/services/public-api-service-typescript/test/public-api-service.spec.ts
```diff
@@ -24,7 +24,11 @@ const transaction: types.Transaction = {
     sender: bs58DecodeRawAddress(senderAddressBase58),
     contractAddress: bs58DecodeRawAddress(contractAddressBase58)
   },
-  payload
+  payload,
+  signatureData: {
```
jlevison: again, just checking that this is intentional

projects/services/public-api-service-typescript/src/service.ts
idandash: Yes

projects/libs/core-library-typescript/src/transaction-pool/transaction-validator.ts
```diff
@@ -31,19 +44,27 @@ export class TransactionValidator {
...
+    const txHelper = new TransactionHelper(tx);
+    const hash = txHelper.calculateHash();
+    const key = ec.keyFromPublic(tx.signatureData.publicKey.toString("hex"));
+    return key.verify(hash, tx.signatureData.signature.toString("hex"));
```
lbeder: Very small suggestion: `verify` can throw, which might crash the service, so perhaps it'd be better to wrap it via `try/catch`.

projects/libs/core-library-typescript/src/transaction-pool/transaction-validator.ts
```diff
@@ -31,19 +44,27 @@ export class TransactionValidator {
...
+  }
+
   private getSubscriptionKey(contractAddress: Address): string {
     // we create a subscription key by zero-left-padding the contract's vchain ID
     const subscriptionKey = "0x0000000000000000000000000000000000000000000000000000000000" + contractAddress.virtualChainId;
     return subscriptionKey;
   }
 
-  constructor(subscriptionManager: types.SubscriptionManagerClient) {
+  constructor(subscriptionManager: types.SubscriptionManagerClient, options: TransactionValidatorOptions = {verifySignature: false}) {
```
lbeder: Shouldn't it be `true` here as well?

PR #163, opened 2018-05-14
Transaction signature - Server side + Javascript client SDK

idandash:


client/client-sdk-javascript/package.json
```diff
@@ -19,14 +19,20 @@
     "@types/chai": "^4.1.2",
     "@types/chai-as-promised": "^7.1.0",
     "@types/express": "^4.11.1",
+    "@types/json-stable-stringify": "^1.0.32",
```
jlevison: all of these additions seem to be some merge issue, is the client-sdk-js really using these?
probably more can removed now that the mock-server is completely gone

client/client-sdk-javascript/src/types/get-port.d.ts
```diff
@@ -0,0 +1,10 @@
+/** Declaration file generated by dts-gen */
```
jlevison: is this used?

e2e/src/simple-smart-contracts.spec.ts
```diff
@@ -13,18 +13,19 @@ const expect = chai.expect;
 chai.should();
 chai.use(ChaiBarsPlugin);
 
-const generateAddress = (): Address => {
+const generateAddress = (keyPair: ED25519Key): Address => {
   const key = new ED25519Key();
```
jlevison: so is key still used?

projects/services/public-api-service-typescript/.gitignore
```diff
@@ -1,2 +1,3 @@
 # Ignore built ts files
 dist/**/*
+src/orbs-api-interface.ts
```
jlevison: i think this is wrong as well

client/client-contract-test/test/client-sdk.spec.ts
```diff
@@ -143,14 +142,20 @@ before((done) => {
 });
 
 describe("The Javascript SDK", () => {
-  const client = new OrbsClient(API_ENDPOINT, SENDER_ADDRESS, TIMEOUT);
-  const contract = new OrbsContract(client, CONTRACT_NAME);
+  const keyPairStub = stubObject<ED25519Key>(new ED25519Key(), ["sign"]);
```
netoneko: I think it would be better to initialize a real instance of `ED2551Key` with a private/public key because you'd have to do it in Java code anyway.

client/client-sdk-javascript/src/orbs-client.ts
```diff
@@ -2,16 +2,21 @@ import { delay } from "bluebird";
...
+import * as stringify from "json-stable-stringify";
 
 export class OrbsClient {
   private endpoint: string;
   readonly senderAddress: Address;
   private sendTransactionTimeoutMs: number;
+  private keyPair: ED25519Key;
 
-  constructor(endpoint: string, senderAddress: Address, sendTransactionTimeoutMs = 5000) {
+  constructor(endpoint: string, senderAddress: Address, keyPair: ED25519Key, sendTransactionTimeoutMs = 5000) {
```
netoneko: It's unclear what `keyPair` stands for: who owns it and why?

client/client-sdk-javascript/package.json
idandash: I think that only Express is not used anymore

PR #164, opened 2018-05-14
Update log size in Winston, use external disk for db and logs

netoneko:
screen shot 2018-05-14 at 11 35 43


projects/libs/core-library-typescript/src/common-library/logger.ts
```diff
@@ -15,7 +15,7 @@ export interface LogzIoOptions {
 export interface FileOptions {
   fileName: string;
   json?: boolean;
-  maxSize?: number;
+  maxsize?: number;
```
lbeder: This is our object, which we later translate to Winston's weird key names, so it should be according to our naming conventions (`maxSize` and not `maxsize`).

netoneko: Fixed.

PR #166, opened 2018-05-14
Implement optional block signature

netoneko:
This is work in progress and currently does not touch block storage configuration, etc.

@idandash please let me know if there is anything I should change before I proceed further.


projects/libs/core-library-typescript/src/common-library/block-utils.ts
```diff
@@ -14,23 +20,60 @@ export namespace BlockUtils {
     return hash.digest();
   }
 
-  export function buildBlock(blockComponents: {header: types.BlockHeader, body: types.BlockBody}): types.Block {
+  export function buildBlock(blockComponents: {header: types.BlockHeader, body: types.BlockBody}, options?: BlockUtilsConfig): types.Block {
```
idandash: I think it's wrong to pass all this config to BlockUtils.  It's enough, IMO, to add an optional key class that can be used to sign a block

idandash: I think it's wrong to pass all this config to BlockUtils.  It's enough, IMO, to add an optional key class that can be used to sign a block

projects/libs/core-library-typescript/test/common-library/block-utils.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+      keyManager,
+      nodeName
+    });
+
+    expect(block).to.have.property("signatureData");
+
+    expect(block.signatureData).to.have.property("signatory", nodeName);
+
+    expect(block.signatureData).to.have.property("signature");
+    expect(block.signatureData.signature).not.to.be.empty;
```
idandash: I think that the test should cover the correctness of the signature, even by stubbing sign()

projects/services/storage-service-typescript/src/server.ts
```diff
@@ -15,14 +16,22 @@ export default function(nodeTopology: any, env: any) {
     const peers = topologyPeers(nodeTopology.peers);
     // TODO: this is very fragile - refactor to make this less hardcoded
     const blockStorageDBPath = BLOCK_STORAGE_DB_PATH || path.resolve(path.join("../../../db", NODE_ENV || "development") + "blocks.db");
+
+    const verifySignature = _.lowerCase(BLOCK_STORAGE_VERIFY_SIGNATURES) === "true";
```
idandash: I think that the default should be that signatures are verified, and they can be disabled under certain circumstances

netoneko: The verification is enabled during deployment, but if we enable it by default we'll have to do additional work around e2e.

PR #167, opened 2018-05-15
mock http server  now validates stubs are only for known endpoints

electricmonk:


netoneko:
@electricmonk please fix tests.

electricmonk:
@netoneko which tests are you referring to? I see that block storage is failing to start but I don't see how my changes could be related to that

PR #172, opened 2018-05-16
Fix pending transaction pool validation

netoneko:


projects/libs/core-library-typescript/test/transaction-pool/pending-transaction-pool.spec.ts
```diff
@@ -137,10 +137,10 @@ describe("Transaction Pool", () => {
         receipt: undefined
       });
     });
-    it("that is in the pending pool", () => {
+    it("that is in the pending pool", async () => {
       const transaction = aValidTransaction();
       const txid = new TransactionHelper(transaction).calculateTransactionId();
-      transactionPool.addNewPendingTransaction(transaction);
+      await transactionPool.addNewPendingTransaction(transaction);
```
netoneko: @idandash that's a touch of a true master.

idandash: LOL

PR #177, opened 2018-05-17
Verify input parameters in JNI

lbeder:


client/crypto-sdk-android/crypto-sdk/src/test/java/com/orbs/cryptosdk/ED25519KeyUnitTest.java
```diff
@@ -85,4 +99,31 @@ public void signs_and_verifies_messages() {
...
+
+    @Test
+    public void throws_on_invalid_arguments_signs_and_verifies_messages() {
+        try (ED25519Key key1 = new ED25519Key();
+             ED25519Key key2 = new ED25519Key()) {
+            byte[] message1 = "Hello World!".getBytes(StandardCharsets.UTF_8);
+
+             try {
+                key1.sign(null);
+            } catch (Exception e) {
```
jlevison: shouldn't there be a assert here in case it does not throw an exception (to fail the test)

jlevison: meaning after the key1.sign()

lbeder: Hmmm... true. I'll improve.

PR #181, opened 2018-05-21
Fix example bot, add docker image for the bot

netoneko:
Adds docker image for the bot so we could deploy it on AWS and see a live demo.

Keys are stored locally in Redis.


client/examples/slack-bot/src/index.ts
```diff
@@ -59,6 +73,16 @@ function mention(client: FooBarAccount) {
   return `<@${client.username}> ${client.address}`;
 }
 
+const redisClient: any = redis.createClient(process.env.REDIS_URL);
+
+async function saveAccount(username: any, keyPair: any) {
+  return redisClient.hmsetAsync(username, { publicKey: keyPair.publicKey, privateKey: keyPair.getPrivateKeyUnsafe() });
```
jlevison: do we need to time limit the objects, maybe its not important but what purges data?

netoneko: There is no need to purge data because the keys are supposed to be stored forever.

PR #183, opened 2018-05-21
Subscription Manager - Adjusted for new ERC20 interface + added rate management

idandash:


projects/libs/core-library-typescript/src/subscription-manager/subscription-manager.ts
```diff
@@ -16,7 +26,58 @@ export class SubscriptionManager {
...
+    if (rateData == undefined) {
+      logger.info(`Could not find rate settings for subscription profile ${subscriptionData.profile}`);
+      return false;
+    }
+
+    const beginningOfThisMonth = new Date(`${now.toISOString().substr(0, 7)}-01`).getTime();
+
+    let minimumRequiredFee: number;
+
+    if (subscriptionData.startTime <= beginningOfThisMonth) {
```
erankirsh: Should test that startTime is in the past. (Not critical because the contract handles it already)

projects/libs/core-library-typescript/src/subscription-manager/subscription-manager.ts
```diff
@@ -16,7 +26,58 @@ export class SubscriptionManager {
...
+      const numOfDaysThisMonth = new Date(now.getFullYear(), now.getMonth() + 1, 0).getDate();
+      const endOfThisMonth = new Date(`${now.toISOString().substr(0, 7)}-${numOfDaysThisMonth}`).getTime();
+
+      const subscriptionDurationInMilliseconds = endOfThisMonth - subscriptionData.startTime;
+      const numOfMillisecondsThisMonth = endOfThisMonth - beginningOfThisMonth;
+
+      minimumRequiredFee = rateData.rate * (subscriptionDurationInMilliseconds / numOfMillisecondsThisMonth);
+    }
+
+    if (subscriptionData.tokens < minimumRequiredFee) {
```
erankirsh: Should probably be <=

projects/services/consensus-service-typescript/src/subscription-manager-service.ts
```diff
@@ -22,7 +22,7 @@ export default class SubscriptionManagerService extends Service {
 
   @Service.RPCMethod
   async getSubscriptionStatus(rpc: types.GetSubscriptionStatusContext) {
```
erankirsh: Should rename to isSubscriptionValid

projects/libs/core-library-typescript/src/subscription-manager/subscription-manager.ts
netoneko: It's the other way around, if we have less than X tokens we throw an error.

PR #184, opened 2018-05-21
vm null args fix +client changes +contract test +android bugfix with url

jlevison:
These are two branches i had combined as when fixing the null args in android i did not want to do it on older code which was not merged yet.

it has:
1. VM will now output a correct error message if args are undefined for any reason
2. Contract test is now more 'programmable' with payload as parameters for the contract
3. Added a test for 'undefined' params by clients in the contract test
4. fixed JS client to support undefined args
5. fixed Android client to support undefined args
6. url builder fix for android client
7. minor refactors+tests for android client

also for reasons i cannot understand, the android docs written last week are somehow in this branch now, no idea how that happened... but they will be pushed now, i will update them later in a different branch.

### DIFF
--- a/client/crypto-sdk-android/readme.md
+++ b/client/crypto-sdk-android/readme.md
@@ -83,7 +83,8 @@ You can create an `OrbsClient` class to call the contract methods with some clie
 
 * `apiEndpoint` - the orbs-network node
 * `senderAddress` - the address of the entity 'acting' right now
-* `timeoutInMs` - optional network timeout, the default is 2000mstimeoutInMs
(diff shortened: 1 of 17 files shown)


client/crypto-sdk-android/readme.md
```diff
@@ -0,0 +1,126 @@
...
+## Integrating the SDK
+
+The package is available via `jcenter()` for download, add to your gradle file the following:
+```gradle
+dependencies {
+    implementation 'com.orbs.sdk.cryptosdk:+'
+}
+```
+
+Currently the latest release version is `0.0.6`
```
netoneko: We won't be able to keep up with the changes.

PR #185, opened 2018-05-22
kinatn will no longer auto-fund accounts

jlevison:


netoneko:
@jlevison wouldn't it impede our progress in testing Kin android integration? I assume transfer transactions won't be processed after this change.

PR #187, opened 2018-05-22
Feature/vm startup check

IdoZilberberg:
Startup check for Virtual Machine Service. It is rather naïve, so any suggestions are welcome.


projects/services/virtual-machine-service-typescript/src/server.ts
```diff
@@ -1,32 +1,37 @@
...
+  const contractRegistryConfig = {
+    contracts: parseContractList(SMART_CONTRACTS_TO_LOAD)
+  };
 
-    const contractRegistryConfig = {
-        contracts: parseContractList(SMART_CONTRACTS_TO_LOAD)
-    };
+  const virtualMachine = new VirtualMachine(contractRegistryConfig, peers.stateStorage);
+  const virtualMachineService = new VirtualMachineService(virtualMachine, nodeConfig);
+  const startupCheckRunner = new StartupCheckRunner("storage", [virtualMachineService]);
```
jlevison: maybe vm?

PR #191, opened 2018-05-23
updated improved raft version

gadcl:


netoneko:
The God of Consensus has committed!

PR #192, opened 2018-05-23
removed the output of keyManager object from logs

jlevison:


netoneko:
I think Winston had something specifically for this: you can pass a list of keys that it will mask automatically.

jlevison:
we use json.stringify there, couldn't find a simple way to do with using a winston filter

PR #196, opened 2018-05-24
Feature/public api startup check

IdoZilberberg:
e2e now starts immediately when all docker containers are healthy; added startup check for Public API Service.


e2e/src/test-environment/orbs-nodes.ts
```diff
@@ -49,7 +49,7 @@ export class OrbsNode implements TestComponent {
       command += " --forceRecreate";
     }
     await this.runDockerCompose(command);
-    await delay(60000);
+    // await delay(60000);
```
jlevison: why was this removed? if not required remove comment

projects/services/public-api-service-typescript/src/service.ts
```diff
@@ -43,6 +52,25 @@ export default class PublicApiHTTPService extends Service {
...
+      startupCheckRunner.run()
+        .then((result: StartupStatus) => {
+          const httpCode = result.status === STARTUP_STATUS.OK ? 200 : 503;
+          return res.status(httpCode).send(result);
+        });
+    });
+
+    const mgmtServer = mgmtApp.listen(httpManagementPort);
+    logger.info(`Started HTTP public api management server on port ${httpManagementPort}`);
+    return mgmtServer;
```
jlevison: it checks that the public api 'business logic' is running, but can it also check if the express started okay and running? (meaning the service itself)

e2e/src/test-environment/orbs-nodes.ts
IdoZilberberg: Replaced by actual logic that proceeds with the test once all docker containers are healthy, instead of waiting arbitrary time. Comment removed.

projects/services/public-api-service-typescript/src/service.ts
IdoZilberberg: Let's talk on Sunday about it. Maybe it's worth adding to the other services too.

e2e/docker-containers-health-check.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
```
netoneko: Is there any reason to use this script instead of writing it in TypeScript? With TS it would be much more maintainable.

projects/services/public-api-service-typescript/src/service.ts
IdoZilberberg: Added check that Public API HTTP Server is alive and listening.

e2e/docker-containers-health-check.sh
IdoZilberberg: It's not the concern of the TS code; also easier to test this code from command line when it's a shell script.

IdoZilberberg: Replaced shell script code with ts code for docker healthcheck.

PR #200, opened 2018-05-24
Added note about Docker tests to e2e/README

IdoZilberberg:


IdoZilberberg:
Changed wording to suggest running test-from-docker.sh on build errors, and mentioned it is slower to run.

PR #203, opened 2018-05-27
Expose network id through configuration

netoneko:


.circleci/e2e-testnet.sh
```diff
@@ -3,5 +3,6 @@
 docker run --rm -ti \
     -e E2E_NO_DEPLOY=true \
     -e E2E_PUBLIC_API_ENDPOINT=$E2E_PUBLIC_API_ENDPOINT \
+    -e NETWORK_ID=54 \
```
jlevison: it becomes a string or number when the code parses it?

netoneko: It's always a string.

PR #204, opened 2018-05-27
KinATN contract argument validations

jlevison:


projects/libs/core-library-typescript/src/virtual-machine/hard-coded-contracts/registry/kinatn-smart-contract.ts
```diff
@@ -8,17 +8,31 @@ export default class KinAtnSmartContract extends BaseSmartContract {
   static readonly POOL_INITIAL_VALUE = KinAtnSmartContract.ONE_BILLION * 20;
 
   public async transfer(recipient: string, amount: number) {
+    if (typeof recipient != "string") {
```
netoneko: I'd use lodash for this, but let's keep it this way.

PR #210, opened 2018-05-28
Feature/sidechain startup check

IdoZilberberg:
Added startup check for Sidechain connector - it only checks the Ethereum connector is defined so comment if you see better alternatives.


IdoZilberberg:
Closing because all its code was merged to master as part of refactor/extract-startup-check-tester branch.

PR #211, opened 2018-05-28
allow 1 node to be out of sync before terminating stub consensus

erankirsh:


projects/libs/core-library-typescript/src/consensus/stub-consensus.ts
```diff
@@ -73,20 +73,25 @@ export class StubConsensus extends BaseConsensus {
       logger.debug(`Stub consensus leader heartbeat - my last block is ${JSON.stringify(myLastCommittedBlock)}`);
 
       // check if all other nodes are synchronized with me
-      let areAllOtherNodesSyncedWithMe = true;
-      if (this.lastBlockHeightByNodeName.size != this.config.clusterSize - 1) {
+      let areEnoughNodesSyncedWithMe = true;
```
netoneko: I think it would be a bit easier for us to maintain this code if we rewrote it using `filter((k, v) => v <= lastBlockSize)` and then check the size.

projects/services/consensus-service-typescript/src/consensus-server.ts
```diff
@@ -96,6 +97,7 @@ export default function(nodeTopology: any, env: any) {
     }
     consensusConfig.leaderNodeName = CONSENSUS_LEADER_NODE_NAME;
     consensusConfig.heartbeatInterval = 1000; // this is the block interval
+    consensusConfig.acceptableUnsyncedNodes = 1; // this is the number of nodes that can be out of sync before stopping the consensus
```
netoneko: Maybe we should expose it as an environment variable.

PR #212, opened 2018-05-28
fixed binary output of contract address to logs

jlevison:


projects/libs/core-library-typescript/src/state-storage/kvstore.ts
```diff
@@ -39,7 +39,7 @@ export class InMemoryKVStore {
   setSync(contractAddress: Buffer, key: string, value: string) {
     const mapKey = this.generateMapKeyString(contractAddress, key);
     this.storeMap.set(mapKey, value);
-    logger.debug(`stored ${contractAddress}.${key} with value ${value}`);
+    logger.debug(`Stored ${contractAddress.toString("hex")}.${key} with value ${value}`);
```
netoneko: I think it should be base58

jlevison: correct, but the store logic is as outputted, see what i changed it to

PR #218, opened 2018-05-29
removed stringify from error logs

jlevison:


projects/libs/core-library-typescript/src/block-storage/block-storage-sync.ts
```diff
@@ -41,7 +41,7 @@ export class BlockStorageSync {
       try {
         await this.blockStorage.addBlock(block);
       } catch (e) {
-        logger.error(`Failed to addBlock() with error: ${JSON.stringify(e)}`);
+        logger.error(`Failed to addBlock() with error: ${e}`);
```
IdoZilberberg: You could also use: logger.error(`Failed to addBlock() with error:`, e);

PR #221, opened 2018-05-30
Gossip Startup Check

IdoZilberberg:


projects/libs/core-library-typescript/src/gossip/gossip.ts
```diff
@@ -162,4 +163,14 @@ export class Gossip {
   public activePeers() {
     return this.clients.keys();
   }
+
+  public async startupCheck(): Promise<StartupStatus> {
+
+    const badClients = _.filter(this.clients || [], (client: WebSocket) => { return !client.readyState || client.readyState !== WebSocket.OPEN; });
+
+    if (badClients.length > 0) {
+      return { name: this.SERVICE_NAME, status: STARTUP_STATUS.FAIL, message: `Found ${badClients.length} bad clients` };
```
netoneko: Even if some of the nodes are unreachable it does not mean that Gossip is not operational. We should return `STARTUP_STATUS.FAIL` only if we don't have any working clients.

PR #224, opened 2018-05-30
Consensus startup check

IdoZilberberg:


projects/libs/core-library-typescript/src/consensus/consensus.ts
```diff
@@ -35,4 +38,15 @@ export class Consensus {
...
   }
+
+  public async startupCheck(): Promise<StartupStatus> {
+    if (!this.actualConsensus) {
+      return { name: this.COMPONENT_NAME, status: STARTUP_STATUS.FAIL, message: "Missing actualConsensus" };
+    }
+
+    return { name: this.COMPONENT_NAME, status: STARTUP_STATUS.OK };
+
+    // return this.consensus.startupCheck();
```
netoneko: Please remove comments.

IdoZilberberg: Removed.

PR #231, opened 2018-06-13
Sync in batches

netoneko:


projects/services/storage-service-typescript/src/block-storage-service.ts
```diff
@@ -170,25 +167,32 @@ export default class BlockStorageService extends Service implements StartupCheck
   }
 
   async onSendNewBlocks(fromAddress: string, payload: any) {
-    logger.info(`Block storage ${this.config.nodeName} received request for new blocks from ${fromAddress}`);
+    logger.info(`Block storage ${this.config.nodeName} received a request for new blocks from ${fromAddress}`);
 
     const blocks = await this.blockStorage.getBlocks(payload.blockHeight);
```
jlevison: this may be a problem in the future, it gets all the blocks and later slices them (line 175)
@erankirsh  - should we invest more time here, to refactor it?

PR #237, opened 2018-07-04
Add scrolling to BlockStorage.getBlocks()

netoneko:


projects/libs/core-library-typescript/src/consensus/benchmark-consensus.ts
```diff
@@ -187,29 +187,29 @@ export class BenchmarkConsensus extends BaseConsensus {
...
-              // logger.info(`Node ${this.node.id} Current benchmark log: `);
-              for (let i = 0; i < benchmarkLog.length; i++) {
-                const blockHash = benchmarkLog[i].term;
-                // BlockUtils.calculateBlockHash(benchmarkLog[i]).toString("hex");
-                rafLogStr += "Height " + i + ":  term:" + benchmarkLog[i].term + " data: " + benchmarkLog[i].data + "\n";
-                // logger.info(`Height ${i}:  ${blockHash}`);
-              }
-              logger.info(rafLogStr);
-            }
+            // const benchmarkLog: any[] = this.node.getLog();
```
jlevison: just add a reason for why this is commented out

netoneko: Done.

PR #238, opened 2018-07-11
corrected signature validation

jlevison:


projects/libs/core-library-typescript/src/common-library/address.ts
```diff
@@ -77,7 +77,7 @@ export class Address {
       // if public key for given, we validate that it's mapped to the right account ID
       address = new Address(publicKey, vchainId, networkId);
       if (!address.accountId.equals(accountId)) {
-        logger.info(`accountId in address and accountId derived by publicKey don't match ${accountId} != ${address.accountId}`);
+        logger.info(`accountId in address and accountId derived by publicKey don't match ${accountId.toJSON().data} != ${address.accountId.toJSON().data}`);
```
netoneko: Derived from public key

